PR #2, opened 2021-09-29
Only copy db for sync, nightly. Files are downloaded on the fly during migration

johangant:


.platform.app.yaml
```diff
@@ -147,12 +147,8 @@ web:
         '^/sites/default/files/(css|js)':
           expires: 2w
 
-# Set up migration content refresh every night with 30 minute gap between
-# file sync and db restoration on this app.
+# Set up migration content refresh every night.
 crons:
-  d7files:
-    spec: '30 0 * * *'
-    cmd: 'migrate-scripts/sync-files.sh $D7_PROJECT_ID $D7_PROJECT_ENV'
```
johangant, line 155: DItched as we're going to download the files we need during the migration itself, from the publicly accessible paths... eg: `https://www.northernireland.gov.uk/sites/default/files/publications/some-doc.pdf`

This saves a whole lot of disk space by not needing us to copy the files from the legacy project into the file mount which migrate would then move/copy the file from into the D9 `public://` namespace/storage area.

.platform.app.yaml
```diff
@@ -147,12 +147,8 @@ web:
...
   d7db:
     spec: '0 1 * * *'
```
johangant, line 153: The db fetch/unpack process takes between 20-30 minutes. This is set to run once a day, but clearly we can do it more often if we need to in future.

config/sync/migrate_plus.migration.d7_file.yml
```diff
@@ -17,7 +15,7 @@ source:
   scheme: public
   batch_size: 1000
   constants:
-    source_base_path: /app/imports/files
+    source_base_path: 'https://www.northernireland.gov.uk/sites/default/files/'
```
johangant, line 18: One currently active D7 hostname. In fact, it might be prudent to always check this is a D7 site in future otherwise you might end up trying to copy files from yourself... depending on how DNS/domains are set up within platform.

config/sync/migrate_plus.migration.d7_file_image.yml
```diff
@@ -25,7 +23,7 @@ process:
     plugin: default_value
     source: language
     default_value: und
-  name: filename
+  name: field_file_image_title_text/0/value
```
johangant, line 26: This is a mandatory field in D7 so I think we can rely on it.

omahm (approved):
LG

PR #3, opened 2021-10-04
Add migrate script; drop files sync script

johangant:
Expecting to add to this over time: aliases, redirects etc to follow.

NB: MySQL timeouts on prod cluster mean long running migrations like the files one may need to be batch executed if running from scratch. Incremental updates should be fine.


DuttonMa (approved):
Looks good

PR #7, opened 2021-11-16
Dept edge build

johangant:
Mostly a rehash/tidy of what's been used before. Needs to execute overnight to help debug any issues with SSH keys or other minor things in there.

The PSH environment is active and up, so if all goes to plan this should build + push back to GH, which in turn triggers the webhook there which prompts PSH to fetch and build/deploy on there.


omahm (approved):
I like the crontab comments 👍

PR #11, opened 2021-11-19
All things around functional tests and services

johangant:
Should let us run tests locally, and scheduled jobs to run against the edge environment.


omahm (approved):
LGTM

PR #13, opened 2021-11-30
Devtools

omahm:
- Option to update Domain hostnames to use the Lando proxy URL's
- Provides an admin toolbar link with list of domain links.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -1,0 +1,75 @@
+<img alt="Drupal Logo" src="https://www.drupal.org/files/Wordmark_blue_RGB.png" height="60px">
+
+Drupal is an open source content management platform supporting a variety of
+websites ranging from personal weblogs to large community-driven websites. For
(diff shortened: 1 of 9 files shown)


johangant:
Does this replicate, or extend, what https://github.com/dof-dss/nicsdru_dept/blob/development/web/sites/default/settings.domain.php aims to do? Granted it's not terribly visible because config overrides aren't in Drupal 8/9.

Good idea with improving visibility of the site URLs

omahm:
> Does this replicate, or extend, what https://github.com/dof-dss/nicsdru_dept/blob/development/web/sites/default/settings.domain.php aims to do? Granted it's not terribly visible because config overrides aren't in Drupal 8/9.
> 
> Good idea with improving visibility of the site URLs

Looks to be fulfilling the same function, just my local instance was using the RL domains which was a pain

johangant (approved):
Let's get this in and see how it behaves as we get into the next dev iteration

PR #20, opened 2021-12-07
Dept 48

omahm:
- Don't display the 'Publish to' option if the user has only 1 Group/Dept membership, warn if there are no memberships
- Fixed bug when publishing new content.


johangant (approved):
LGTM

PR #21, opened 2021-12-08
Easychart node type, config and migration config

johangant:
Initial config - needs refinement later on for other supporting fields and mappings, eg: taxonomy terms


omahm (approved):
LG, nice one

PR #23, opened 2021-12-09
Dept 50

omahm:
Enabled a bunch of modules installed on the D7 version.


composer.json
```diff
@@ -83,6 +99,7 @@
         "drupal/migrate_upgrade": "^3.2",
         "drupal/moderation_sidebar": "^1.4",
         "drupal/noreferrer": "^1.10",
+        "drupal/password_policy": "^3.0",
```
johangant, line 102: So we took this out of NIDirect because it caused some problems (can't recall what, specifically) and replaced with a more basic, custom code module (nidirect_password_policy). Perhaps we can convert that module into an origins one that can be re-used across any site we operate?

johangant (commented):
All good, but I'd consider moving the nidirect_password_policy module into an origins one as password_policy caused us problems in the past.

Perhaps create a new origins_password_policy module first, then we can retired the nidirect one in a later release on that project.

PR #31, opened 2021-12-21
Migrate lookup plugins for D7 entities, updated config to use them

johangant:


config/sync/migrate_plus.migration.group_node_easychart.yml
```diff
@@ -56,11 +40,14 @@ process:
       plugin: migration_lookup
       migration: node_easychart
       source: uuid
-  uid: uid
+  uid:
+    -
+      plugin: d7_user_lookup
+      source: node_uid
```
johangant, line 46: Easy to miss this - need to check what the field ids are in the source plugin when this happens. I kept getting NULL for uid and was very confused for a while.

config/sync/migrate_plus.migration.users.yml
```diff
@@ -0,0 +1,54 @@
+uuid: 5a897396-9210-4c79-b877-2e3107ded0c4
+langcode: en
+status: true
+dependencies: {  }
+id: users
```
johangant, line 5: Couldn't recycle the d7_users.yml core migration, as it griped about an incompatible core version (??). This works well and means the migrations are also neatly tagged/grouped and all the d6/d7 upgrade cruft can be ignored/filtered out.

web/modules/custom/dept_migrate/modules/dept_migrate_group_nodes/config/install/migrate_plus.migration.group_node_easychart.yml
```diff
@@ -22,22 +22,23 @@ process:
...
-        - 'infrastructure'
-        - 'dcal'
-        - 'doe'
-        - 'justice'
-        - 'communities'
+#    -
+#      plugin: skip_on_value
+#      method: row
+#      source: domain_user/0/target_id
+#      # Whenever a site goes live (prod in D9), add its key to this list.
```
johangant, line 29: The default state we'll begin with is to migrate everything for all sites. When a site launches to D9, prod content entry will happen there instead of D7 so we can begin to ignore things from that domain id from there on. The migrations will get faster as a result of having to process less material.

web/modules/custom/dept_migrate/src/MigrateUuidLookupManager.php
```diff
@@ -139,6 +140,55 @@ public function lookupBySourceNodeId(array $nids) {
...
+  /**
+   * @param array $uids
+   *   One or more source user ids.
+   *
+   * @return array
+   *   Associative array of user metdata, keyed by source user id
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  public function lookupBySourceUserId(array $uids) {
```
johangant, line 152: I get the feeling these can be reworked to be more general and by passing in the entity type and ids needed. That said, they're working for now, so I'm minded to carry on with a more verbose but more literal approach on these. This will be a throwaway class for supporting the migrations so efficiency isn't a huge priority right now.

web/modules/custom/dept_migrate/src/Plugin/migrate/process/D7UserLookup.php
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function transform($value, MigrateExecutableInterface $migrate_executable, Row $row, $destination_property) {
+    // Cast the $value to be an array, even for single values.
+    if (!is_array($value)) {
+      $value = (array) $value;
+    }
+
+    $user_metadata = $this->lookupManager->lookupBySourceUserId($value);
```
johangant, line 70: This plugin acts as a wrapper around the service class and lets us have a more concise migration process plugin chain. In some cases you can work around this with multiple migrate lookups but often, like for entity reference field values, you can't do this as your input is a D7 entity id rather than a UUID.

web/modules/custom/dept_user/src/Plugin/migrate/source/d7/D7UserUuid.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    return parent::prepareRow($row);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getIds() {
+    $ids['uuid']['type'] = 'string';
```
johangant, line 97: This change means the entire migration needs uninstalling and the db tables dropped, so that when re-run it creates them again with the correct structure.

omahm (approved):
Nice one and great commenting

PR #35, opened 2022-01-07
Fix broken group video plugin and config

johangant:


config/sync/migrate_plus.migration.group_media_video.yml
```diff
@@ -19,13 +19,20 @@ source:
...
-    default_value: group_content_type_7d15138adb66b
+    default_value: group_content_type_770174f09a085
   gid:
     -
       plugin: default_value
       default_value: 1
-  label: title
-  entity_id: fid
+  label: filename
+  entity_id:
```
johangant, line 28: If the d7_file_media_video migration's VideoUrl.php migrate process plugin fails to fetch a thumbnail image, it used to ignore the item and skip on. This causes a problem because this migration uses the D7 source as the guide, then tries to create group_content relation entities... and if we've skipped a video entity we can't then load it in D9 to create the relation... you get a big nasty group or contententitybase error. That's quite obscure to debug, too.

web/modules/custom/dept_migrate/modules/dept_migrate_group_files/config/install/migrate_plus.migration.d7_file_media_video.yml
```diff
@@ -16,7 +16,6 @@ source:
   constants:
     bundle: 'remote_video'
 process:
-  mid: fid
```
johangant, line 19: Got rid because we want to rely on the D9 autoincrement, rather than carry these across verbatim.

web/modules/custom/dept_migrate/modules/dept_migrate_group_files/config/install/migrate_plus.migration.group_media_video.yml
```diff
@@ -15,13 +15,20 @@ source:
 process:
   type:
     plugin: default_value
-    default_value: group_content_type_7d15138adb66b
+    default_value: group_content_type_770174f09a085
```
johangant, line 18: Seems to have changed. Odd, but shouldn't need to change anymore in future?

web/modules/custom/dept_migrate/modules/dept_migrate_group_files/src/Plugin/migrate/process/VideoUrl.php
```diff
@@ -25,24 +25,7 @@ public function transform($value, MigrateExecutableInterface $migrate_executable
     $url = preg_replace('|^oembed:\/\/|', '', $value);
     // Remove HTML special chars to give a user readable string.
     $url = urldecode($url);
-    // See if thumbnail is available.
-    $thumbnail_url = '';
 
-    if (preg_match('|vimeo|', $url)) {
```
johangant, line 31: This doesn't help us here. If a thumbnail's missing, we can live with it. But we can't live without pulling the entity across into D9

PR #37, opened 2022-01-10
DEPT-51 news nodes migration

johangant:


ooneill547 (approved):

johangant:
Would quite like to merge this irrespective of deprecated code in Origins modules, as it'll be useful for tomorrow's show & tell. Mindful that origins needs its own maintenance to keep it up to date, and it'll sort itself out across all projects when that's done.

PR #40, opened 2022-01-12
Deprecated fixes

omahm:
- Bunch of automatic config changes
- Fixed some deprecated warnings
- Added accessibility method to Department class
- Added cache tags to accessibility menu link

PR #43, opened 2022-01-17
Adjust footer menu cache context and tag allocation

johangant:
Broadens the context of the menu so that it generates once per theme, rather than adding duplicate cache entries by user permission/role default).

Also adds cache tags for the group entity id and node id so that if either change, the footer menu is regenerated.


ooneill547 (approved):

johangant:
Merged in https://github.com/dof-dss/nicsdru_dept/pull/40 to consolidate the cache tags for group and accessibility statement node. It now changes if you add/remove the value from the group or if you change the node.

PR #47, opened 2022-01-24
Improved handling of multi-value domain settings from D7

johangant:
* Dropping group_entity migration config as only supported 1-1 mappings.
* Covering the complex group content entity creation/management in eventsubscriber classes, invoked after every migrate row is processed.

Covers nodes and user entities from D7. Media isn't affected as it's not assigned to a domain in D7


web/modules/custom/dept_migrate/modules/dept_migrate_group_nodes/src/EventSubscriber/NodeEntityMigrateSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Callback for post-row-save event.
+   *
+   * @param \Drupal\migrate\Event\MigratePostRowSaveEvent $event
+   *   The migrate post row save event.
+   */
+  public function onPostRowSave(MigratePostRowSaveEvent $event) {
+    if (preg_match('/^node_/', $event->getMigration()->id())) {
```
johangant, line 44: Important to check the migration id, otherwise this callback happens on every row for any migration run... which isn't what we want :)

web/modules/custom/dept_migrate/modules/dept_migrate_group_users/dept_migrate_group_users.services.yml
```diff
@@ -0,0 +1,6 @@
+services:
+  dept_migrate_group_users.eventsubscriber.post:
+    class: 'Drupal\dept_migrate_group_users\EventSubscriber\UserEntityMigrateSubscriber'
```
johangant, line 3: I've split these into two different handlers for clarity over efficiency. It might well be possible to crunch down to a single event subscriber and hand off the migrate row/entity to the required callback based on entity type... but there's probably not a lot in it in terms of maintenance overhead right now. Open to feedback!

web/modules/custom/dept_migrate/src/MigrateSupport.php
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  /**
+   * @param string $domain_machine_name
+   *   The machine name of the D7 domain.
+   * @return int
+   *   The group id (GID) of the group entity.
+   */
+  public function domainToGroupId(string $domain_machine_name) {
```
johangant, line 83: Static map/lookup as to my knowledge we're not routinely adding new sites/domains.

web/modules/custom/dept_migrate/src/MigrateSupport.php
```diff
@@ -0,0 +1,269 @@
...
+   *   The node object.
+   */
+  protected function syncNodeDomainsToGroups(NodeInterface $node) {
+    // Fetch the map/details of the D7 entity.
+    $d7_entity = reset(\Drupal::service('dept_migrate.migrate_uuid_lookup_manager')->lookupByDestinationNodeIds([$node->id()]));
+
+    if (!empty($d7_entity['domains'])) {
+      $plugin_id = 'group_node-' . $node->bundle();
+
+      $group_content_data = $this->dbconn->query("
```
johangant, line 145: So this approach uses a static query to test whether a domain/group has content for this entity or not. We can do the same using Entity API, but it's actually quite verbose and a little less performant. If this code were being used for anything other than migrations, I'd probably opt to use Entity API and do something like load the group content entity, search for content in a given group entity, then add/delete a required. That's quite fiddly, this achieves the same thing and is _slightly_ faster.

web/modules/custom/dept_migrate/src/MigrateSupport.php
```diff
@@ -0,0 +1,269 @@
...
+
+      foreach ($d7_entity['domains'] as $domain_machine_name) {
+        $d7_mapped_group_ids[] = $this->domainToGroupId($domain_machine_name);
+      }
+
+      $d7_mapped_group_ids = array_unique($d7_mapped_group_ids);
+
+      if (empty(array_diff($d7_mapped_group_ids, $node_group_ids))) {
+        // No changes, stop and return here.
+        return;
```
johangant, line 171: Bailing out as early as possible here as we don't want to be deleting/re-adding the same thing each time, only if there are differences between D9 group content and D7 domain assignments.

web/modules/custom/dept_migrate/src/MigrateSupport.php
```diff
@@ -0,0 +1,269 @@
...
+      // From here, we know we need to update group values, so we begin by
+      // removing all known group content values for this entity type and plugin
+      // then re-inserting the group content entities to the groups needed.
+      // Using static queries rather than the entity API for speed.
+      // Entity API equivalent requires multiple stages of group content
+      // entity lookups, empty checks, explicit entity delete commands etc.
+      // Here, we know we have two tables and need to junk it out by entity
+      // id and content/plugin type which we can do with a single SQL query.
+      $this->dbconn->query("
+        DELETE g, gfd
```
johangant, line 183: Don't think this syntax works outside of MySQL/MariaDB but that's what we're using for the foreseeable so.... :)

web/modules/custom/dept_migrate/src/MigrateSupport.php
```diff
@@ -0,0 +1,269 @@
...
+        $group->addContent($node, 'group_node:' . $node->bundle());
+      }
+    }
+  }
+
+  /**
+   * @param \Drupal\user\UserInterface $user
+   *   The user entity.
+   */
+  protected function syncUserDomainsToGroups(UserInterface $user) {
```
johangant, line 205: Again, repetition could be reduced a bit with further refinement but there are enough differences in tables/plugin ids between nodes and users that it's just easier to split out into its own function.

web/modules/custom/dept_migrate/src/MigrateSupport.php
```diff
@@ -0,0 +1,269 @@
...
+      // Compare current groups to current domains, if they differ then
+      // update group content entity values.
+      $d7_mapped_group_ids = [];
+
+      foreach ($d7_entity['domains'] as $domain_machine_name) {
+        $d7_mapped_group_ids[] = $this->domainToGroupId($domain_machine_name);
+      }
+
+      // Dedupe.
+      $d7_mapped_group_ids = array_unique($d7_mapped_group_ids);
```
johangant, line 235: Important, because without it... old domains that are mapped to the same group result it duplicates group content entity creation. De-duping here stops it.

omahm (approved):
All looks good, nice job 👍

PR #48, opened 2022-01-25
Dept 69

omahm:


web/modules/custom/dept_migrate/modules/dept_migrate_group_nodes/config/install/migrate_plus.migration.node_subtopic.yml
```diff
@@ -0,0 +1,89 @@
...
+  field_summary:
+    -
+      plugin: get
+      source: field_summary
+  field_banner_image:
+    -
+      plugin: sub_process
+      source: field_banner_image
+      process:
+        target_id: fid
```
johangant, line 74: This might need updating (see node_news migration) to pull the correct file id based on a lookup plugin. Otherwise it's not guaranteed to be a match as files will be added at both D7 and D9 sides during the progressive migration of site content.

johangant (changes requested):
Looks good, might need to add the logger stuff to other handlers too in case something's missing. One small comment around a fid needing a lookup plugin... think it's present in everything else but missed in one field mapping alone.

web/modules/custom/dept_migrate/modules/dept_migrate_group_nodes/config/install/migrate_plus.migration.node_subtopic.yml
omahm: Correctly did this within topic and forgot about this field, nice catch.

johangant (approved):

PR #50, opened 2022-01-26
Dept 70

omahm:


johangant:
Is this the same outcome as if you'd taken any entity ref migration field and applied a lookup plugin to the value, a bit like this (although replace the file lookup with some kind of node id lookup plugin that calls the migrate_uuid_lookup_manager service):

```
  field_video:
    -
      plugin: sub_process
      source: field_video
      process:
        target_id:
          plugin: d7_file_lookup
          source: fid
 ```
 
 Honestly not fussy if the outcome is the same and performance is comparable :)

omahm:
> Is this the same outcome as if you'd taken any entity ref migration field and applied a lookup plugin to the value, a bit like this (although replace the file lookup with some kind of node id lookup plugin that calls the migrate_uuid_lookup_manager service):
> 
> ```
>   field_video:
>     -
>       plugin: sub_process
>       source: field_video
>       process:
>         target_id:
>           plugin: d7_file_lookup
>           source: fid
> ```
> 
> Honestly not fussy if the outcome is the same and performance is comparable :)

This should be a lot quicker as it performs the lookup against the distinct list of nids versus looking up each individually.

johangant (approved):

PR #51, opened 2022-02-28
Baseline URL aliases migrate config and group

johangant:
Adds two key migrations for URL aliases (aka path aliases): nodes and taxonomy terms. We discount the rest as we don't expect to re-use aliases for managed files, users or media entities - canonical paths should suffice for those.

There's some minor updates for PHPCS which was due to upstream changes in the static analysis libraries supporting that process.


omahm (approved):
LGTM

PR #52, opened 2022-03-03
Dept 74

omahm:


web/modules/custom/dept_migrate/modules/dept_migrate_group_nodes/src/Plugin/migrate/source/d7/NodeUuid.php
```diff
@@ -206,6 +206,7 @@ public function fields() {
    */
   public function getIds() {
     $ids['uuid']['type'] = 'string';
+    $ids['nid']['type'] = 'integer';
```
johangant, line 209: If we're tracking D7 content by node ID, do we need to rename the class to make it clear we don't need to rely on UUID anymore?

omahm: That's a good point, I'll rename it

johangant (approved):

PR #53, opened 2022-03-04
Add redirects migration and row prepare subscriber event

johangant:
Rather than try and wrestle migrate pseudo fields in to break up/rewrite node paths, it's easier to add an event subscriber which can prepare the row before it saves. That can then detect any internal node paths and update the numerical value to the correct D9 version, if it exists.


ooneill547 (approved):

omahm (approved):
Nice one, might be able to make a few adjustments when the new migrations are in place.

PR #55, opened 2022-03-07
Consultations migration

johangant:
Migrate config for daterange


config/sync/migrate_plus.migration.d7_file_media_document.yml
```diff
@@ -0,0 +1,59 @@
...
+    -
+      plugin: get
+      source:
+        - field_file_title
+        - filename
+    -
+      plugin: callback
+      callable: array_filter
+    -
+      plugin: callback
```
johangant, line 36: This odd pipeline enables the following idea: use the title field value but if it's empty for any reason use the filename instead.

web/modules/custom/dept_migrate/modules/dept_migrate_group_files/src/EventSubscriber/MediaEntityMigrateSubscriber.php
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Callback for post-row-save event.
+   *
+   * @param \Drupal\migrate\Event\MigratePostRowSaveEvent $event
+   *   The migrate post row save event.
+   */
+  public function onPostRowSave(MigratePostRowSaveEvent $event) {
+    if (preg_match('/^d7_file_media_/', $event->getMigration()->id())) {
```
johangant, line 59: Rather than use dedicated migrate configs to create the group content/relation entities, this event subscriber adds them for any media migration. Benefits: more control + visibility, and cleaner migrate:status output.

PR #57, opened 2022-03-08
Dept 77

omahm:


johangant (approved):

web/modules/custom/dept_migrate/modules/dept_migrate_group_nodes/src/Plugin/migrate/source/d7/Node.php
```diff
@@ -207,6 +214,7 @@ public function fields() {
   public function getIds() {
     $ids['uuid']['type'] = 'string';
     $ids['nid']['type'] = 'integer';
+    $ids['domains']['type'] = 'string';
```
johangant, line 217: That's so straightforward... I always assumed these were single, PK values. Clue's in the name I suppose... Ids... plural :)

omahm: technically they are supposed to be treated as PK's but it's the only way I could find to extend the mapping schema

PR #60, opened 2022-03-11
Publication content type adjustments, migrate config, some platform + script config

johangant:


.platform.app.yaml
```diff
@@ -152,6 +152,9 @@ crons:
   d7db:
     spec: '0 1 * * *'
     cmd: 'migrate-scripts/sync-db.sh $D7_PROJECT_ID $D7_PROJECT_ENV'
+  sync_files:
+    spec: '30 0 * * *'
+    cmd: 'migrate-scripts/sync-private-files.sh $D7_PROJECT_ID $D7_PROJECT_ENV'
```
johangant, line 157: Need this to periodically fetch the private files from D7, as we can't use the HTTP approach to download them like with public files; Drupal handles each request for private files and we'd otherwise get HTTP 403.

config/sync/migrate_plus.migration.d7_file_private.yml
```diff
@@ -0,0 +1,46 @@
...
+      delimiter: /
+      source:
+        - constants/source_base_path
+        - filepath
+  uri:
+    plugin: file_copy
+    source:
+      - '@source_full_path'
+      - uri
+    move: false
```
johangant, line 33: Need to copy as otherwise we'd end up re-importing files we've already got, as the nightly sync would re-fetch them and they'd be seen as new things to import.

config/sync/migrate_plus.migration.d7_file_private.yml
```diff
@@ -0,0 +1,46 @@
...
+migration_tags:
+  - dept_sites
+migration_group: migrate_group_files
+label: 'Private files'
+source:
+  plugin: d7_file_uuid
+  scheme: private
+  track_changes: true
+  constants:
+    source_base_path: /app/imports/files/private/
```
johangant, line 18: These get copies into /app/private once D9 has registered them as a managed file

ooneill547 (approved):

PR #70, opened 2022-03-17
Dept 95

omahm:


config/sync/google_map_field.settings.yml
```diff
@@ -0,0 +1,5 @@
+_core:
+  default_config_hash: mWL2dHvi9X6xknPeaUo2dSbC4IGZ-o-bdn2kRs0MzII
+google_map_field_auth_method: 1
+google_map_field_apikey: ''
```
johangant, line 4: Do we need a settings.php config override so we can set the value from an env var?

omahm: That's true.

johangant (approved):
LGTM, merge in when ready

PR #77, opened 2022-04-04
Fix document file lookup and config

johangant:


web/modules/custom/dept_migrate/modules/dept_migrate_group_files/config/install/migrate_plus.migration.d7_file_media_document.yml
```diff
@@ -43,6 +43,8 @@ process:
   field_media_file/target_id:
     - plugin: d7_file_lookup
       source: fid
+    - plugin: default_value
+      default_value: 99316
```
johangant, line 47: I don't like magic numbers but I can't see an easy way to make this dynamic. It's impossible to rollback the placeholder document + file unless deleted outside of the migrations (UI or CLI) so this should hold until it's no longer needed.

web/modules/custom/dept_migrate/src/MigrateUuidLookupManager.php
```diff
@@ -325,6 +325,11 @@ public function lookupBySourceFileId(array $fids) {
 
       // Switch table if there's a specified type for this file.
       switch ($file['d7type']) {
+        case 'document':
```
johangant, line 328: This explains a lot about why the file lookups for documents returned strange results. Oops

omahm (approved):
LG

PR #80, opened 2022-04-07
Enable cron task for migrate script. Script refined for better performance

johangant:
Tee command will flush script output to a dedicated file (overwritten each time) so we can review outside of the general app logs.


johangant:
Merging in to review progress overnight

PR #84, opened 2022-04-12
Book module config, collection and book migrate config

johangant:
Importing books and config relies on:

* Book module being enabled
* Collection nodes being imported *first* (they act as the book entity/housing themselves)
* Book migration happening which links together existing nodes (a variety of content types) and allocating them to the Collection nodes from before.

Migrations still suffer from occasional instances where a D7 or D9 node can't be found, for whatever reason. We'll be looking to improve visibility of that in a later feature.

### DIFF
--- a/config/sync/migrate_plus.migration.dept_book.yml
+++ b/config/sync/migrate_plus.migration.dept_book.yml
@@ -1,0 +1,37 @@
+uuid: f72dd208-58b8-48b7-be5b-0972e509b11b
+langcode: en
+status: true
+dependencies: {  }
+id: dept_book
+class: null
+field_plugin_method: null
+cck_plugin_method: null
+migration_tags:
+  - dept_sites
+migration_group: migrate_group_book
+label: 'Departmental books'
+source:
+  plugin: book
+process:
+  nid:
+    -
+      plugin: d7_node_lookup
+      source: nid
+  book/bid:
+    -
+      plugin: d7_node_lookup
+      source: bid
+  book/weight:
+    -
+      plugin: get
+      source: weight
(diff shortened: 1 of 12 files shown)


config/sync/migrate_plus.migration.dept_book.yml
```diff
@@ -0,0 +1,37 @@
...
+field_plugin_method: null
+cck_plugin_method: null
+migration_tags:
+  - dept_sites
+migration_group: migrate_group_book
+label: 'Departmental books'
+source:
+  plugin: book
+process:
+  nid:
```
johangant, line 16: Much lifted from here https://github.com/dof-dss/nidirect-d8-mig-mods/blob/development/migrate_nidirect_global/config/install/migrate_plus.migration.nidirect_book.yml as inspiration, but you'll note in a few key places we have to pass the original nid value into the lookup plugin to get the D9 nid.

config/sync/migrate_plus.migration.node_collection.yml
```diff
@@ -0,0 +1,66 @@
...
+class: Drupal\migrate\Plugin\Migration
+field_plugin_method: alterFieldInstanceMigration
+cck_plugin_method: null
+migration_tags:
+  - dept_sites
+migration_group: migrate_group_nodes
+label: 'Collection content type nodes'
+source:
+  plugin: d7_dept_node
+  node_type: book
```
johangant, line 15: These nodes are book nodes in D7, but with a custom label (Collection)

omahm (approved):
Nice one, LGTM 👍

PR #90, opened 2022-04-25
Dept 114 low level approach

johangant:
A bit rough around the edges but this PR would:

- Provide a service function to rebuild the group content relation entities for a given node type.
- Doesn't remove content for anything that isn't tracked in a migration (useful for running on an already live site) 

For local debugging, you can reset the autoincrement ids on the group_content and group_content_field_data tables. Best not to on prod, or if you need to, everything needs to come out first to avoid problems.


johangant:
Closing off as the sproc approach works as well, in a more concise manner. Will leave the branch remaining just in case we need it in future.

PR #93, opened 2022-05-02
Dept 97

omahm:
- Providing a domain mapping configuration. We will need this when department sites are moved to alternative domains.
- Remove reliance on dept manager service.
- Write out all valid domain mapping when source node has a domain id of 0.


johangant (approved):
LG

PR #99, opened 2022-05-06
Dept 108

omahm:
Initial stab at a migration summary dashboard


johangant (approved):
group_relationships table doesn't seem to create on install, but the rest works well :shipit:

omahm:
> group_relationships table doesn't seem to create on install, but the rest works well :shipit:

It gets removed and rebuilt in the sproc.

PR #103, opened 2022-05-09
DEPT-117 search facets

johangant:


omahm (approved):
LG

ooneill547 (approved):
It might be a good idea to install the Date Facet Drilldown module to allow the date facets to be filtered by month as well as year

johangant:
> It might be a good idea to install the Date Facet Drilldown module to allow the date facets to be filtered by month as well as year

Good spot, I didn't see how it changed when selected. I'll add that now.

PR #104, opened 2022-05-10
Dept 122

omahm:
Replace loop with cursor to fix duplicate issues.


johangant (approved):
Looks excellent, nice one :+1:

PR #107, opened 2022-05-11
Add env var for solr host/port on psh

johangant:
Don't perceive this to be a service disclosure risk, as that information is easily derived from PSH public docs around Solr service defaults for Drupal projects and there's little that can be done to access this service from outside the cluster without an authenticated SSH tunnel. Can also store this in Circle CI but the value is then obfuscated even from admin view which takes longer for future debugging.


omahm (approved):
Yeah, agree with your on this.

PR #109, opened 2022-05-13
Migrate config updates

johangant:
Trim node titles, fix topic mappings, fix format defaults for body field, fix broken media lookups


config/sync/core.extension.yml
```diff
@@ -130,7 +130,6 @@ module:
   responsive_image: 0
   scheduled_transitions: 0
   scheduler: 0
-  scheduler_content_moderation_integration: 0
```
johangant, line 133: Wasn't listed on NIDirect so turned it off here too. Can always re-enable later if needed

omahm (approved):
Good job 👍

PR #115, opened 2022-05-17
Fix body bypass parameter, correct file/image lookups

johangant:


omahm (approved):
More bypasses than an overweight American. 👍

PR #116, opened 2022-05-17
Double solr disk size, enlarge main db from excess in d7 db

johangant:
- Metrics show D7 db size is only at 30% usage - likely due to trimming of unwanted cache tables etc.
- Increase Solr disk space size as we're filling it with much more content now.
- Increase D9 db disk size to accommodate extra content (by recycling the excess from the D7 db service we weren't using)


omahm (approved):
Make it large 🍟

PR #138, opened 2022-06-16
Dept 134: topics listing

johangant:
Confirmed that this only shows topics providing topic nodes aren't in the node_access table with the 'all' realm. Fortunately, etgrm sets these correctly to group_domain_id with the expanded domain id values as required.


johangant:
Agh, wrong branch. Recreating this on the correct branch

PR #140, opened 2022-06-17
Row save event subscriber to sort out node_access values for imported nodes

johangant:
Could potentially make the domain_ids temp table a permanent table (for now), if we have multiple things using it for the foreseeable? For now, it's a private function for use in here as the domains are established + static.


omahm:
Is there a particular reason for using this instead of the sproc that builds the node_access entries as part of the group relationships?
 https://github.com/dof-dss/nicsdru_dept/blob/development/web/modules/custom/dept_migrate/modules/dept_etgrm/inc/sql_update_node_access.sproc

johangant:
> Is there a particular reason for using this instead of the sproc that builds the node_access entries as part of the group relationships? https://github.com/dof-dss/nicsdru_dept/blob/development/web/modules/custom/dept_migrate/modules/dept_etgrm/inc/sql_update_node_access.sproc

It's taken me a while, but having tidied up some stale data and, crucially, executed the fill chain of sprocs rather than just the final one, I can see that the end result corrects the issue that nodes migrated using migrate API end up with the 'all' realm.

Two final problems:

1. If we execute the etgrm:ca command at the end of each nighly migration, our group_content_field_data table's auto increment id values is going to be huge in a short period of time. We can fix that by resetting the id increment value but it would mean junking out and rebuilding the entire table for all entity types to avoid clashes.
2. Ad hoc migrate commands run over the CLI won't be connected to the sproc invocation. It's possible to invoke it from an event subscriber listening to the post migration event and only run it for node migrations, but that puts more churn on the issue above as the etgrm command will be run more often.

PR #144, opened 2022-06-24
DEPT 62: RSS feeds for publications by topics/subtopics

johangant:
* Migrate config change needed to ensure field_site_subtopic and field_site_topics use node lookups: they're entity reference fields looking for nodes, not taxo terms - that's the global topics field.
* Minor query adjustment needed when adding parent group relationship + filter on a view. We need to set it to the current department/group id which can't be done with the present filter options. This + distinct operator ensures that the results returned for anon + auth users are the same, subject to any node access restrictions, and the group_content_field_data records from the relationship doesn't add duplicate rows where a node sits in multiple groups.

### DIFF
--- a/web/modules/custom/dept_node/dept_node.module
+++ b/web/modules/custom/dept_node/dept_node.module
@@ -7,6 +7,7 @@
 
 use Drupal\dept_node\Entity\Node;
 use Drupal\dept_node\Form\NodeForm;
+use Drupal\views\ViewExecutable;
 
 /**
  * Implements hook_entity_type_alter().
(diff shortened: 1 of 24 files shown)


web/modules/custom/dept_node/dept_node.module
```diff
@@ -43,3 +46,45 @@ function dept_node_form_node_form_alter(&$form, $form_state) {
...
+function dept_node_views_pre_render(ViewExecutable $view) {
+  // Append the active site name to a few fields/titles for publications RSS.
+  if ($view->id() === 'press_releases') {
+    $site_name = \Drupal::config('system.site')->get('name');
+    $title = t('Latest press releases from ') . $site_name;
+    $view->setTitle($title);
+
+    if (\Drupal::routeMatch()->getRouteName() === 'entity.view.preview_form') {
+      \Drupal::messenger()
+        ->addWarning('NB: Title alterered in dept_node_views_pre_render()');
```
omahm, line 62: Nice touch.

PR #150, opened 2022-08-04
Dept 170

johangant:
* Main fix was for incorrect params being passed to a migrate event subscriber - looks like some args/defaults have changed in core.
* Minor but extensive config tweaks from core, predominantly uuids and lazy loading attributes for some entity config.


omahm (approved):
👍

ooneill547 (approved):

PR #153, opened 2022-08-08
Dept 171

omahm:


config/sync/core.extension.yml
```diff
@@ -23,7 +23,6 @@ module:
   cookie_content_blocker: 0
   cookie_content_blocker_media: 0
   csp: 0
-  ctools: 0
```
johangant, line 26: Ctools finally going!? Hooray :)

johangant (approved):
Just need the build/composer to resolve itself to a set of packages it can validate

config/sync/core.extension.yml
omahm: Yeah, pathauto was the only module with a dependency and that's now been sorted.

johangant:
> I've removed a some package installs and the custom php config to get it passing. The new Cimg image includes libpng. I'm not sure what 'pcntl' and 'posix' was required for?

Those are for the platform.sh CLI tool

omahm:
> > I've removed a some package installs and the custom php config to get it passing. The new Cimg image includes libpng. I'm not sure what 'pcntl' and 'posix' was required for?
> 
> Those are for the platform.sh CLI tool

Hmm that's going to be a problem

PR #160, opened 2022-08-18
DEPT-175 topics service class

johangant:


omahm (approved):
Nice one. A future addition might eb to return the topic entities themselves.

PR #164, opened 2022-09-01
DEPT-179 topics list on node title area, misc tooling rejigging

johangant:


omahm (approved):
LGTM. Some nice tooling changes in there

PR #205, opened 2022-10-12
DEPT-36 News and press-releases config and custom code

johangant:
So quite a large one, sorry. Some key points:

- NIGOV wants to show news nodes tagged with press releases news type at /press-releases and 404 at /news.
- All other sites want to 404 for /press-releases and show other news at /news
- The view displays powering those routes will by default respond to any group as they're part of the same codebase.

To work around this, we use views' contextual argument handler to return 404 when nothing is provided. In dept_search we intercept the active dept/group id and pass that to an early stage views hook. If it's a valid group id, then it shows the content. Bonus points: we can specify multiple groups, if we want to. NB: you need to take off any built-in Solr processors for content access checks as well as on the views config too or you _might_ end up excluding things by accident.

NB NB: The default domain (NIGOV) should let you view any published node anyway, as domain_group's handler class controls this.

Cache contexts and tags are added for `url.site` so we don't render, for example, DAERA's news page then go to NIGOV expecting a 404 and have Drupal go "oh, this page again" and serve the wrong site's content thinking it's a request for the same page when the site context is different.


ooneill547 (approved):

omahm (approved):
Good job Johan

PR #215, opened 2022-10-20
Dept 207: introduce group 2.x and related adjustments

johangant:


johangant:
Putting this PR on ice for now while further obstacles in supporting contrib modules are overcome

PR #221, opened 2022-10-25
DEPT-224 ~ Update role permissions

omahm:
Not a fully comprehensive update but an update at least. Will require input from NICS staff.


johangant (approved):
I bet you were cross-eyed at the end of this :+1:

omahm:
> I bet you were cross-eyed at the end of this 👍

That and RSI 😢

PR #224, opened 2022-11-02
DEPT-255 turn off content moderation for importing nodes + updates to them

johangant:
Content moderation (CM) doesn't seem to get on with migrations that track and import changes over time. You either end up with a busy and unnecessary revisions list (fixed with patch to provide suitable use of sync attribute in migrate module) or creating new drafts on migrated content creates a jumble of content moderation entity revisions which look very complex to detangle.

Removing any moderation YML, turning off CM and then importing changes (new and updated) works well. You can then turn on CM again and continue to make drafts which then show the correct changes.


omahm (approved):
LGTM

PR #225, opened 2022-11-02
Tweaks to publications search page

neilblair:
Add result heading showing number of publications, trim summaries and strip html, sort by publication date


ooneill547 (approved):

johangant (approved):
Looks good, thanks. One thing I noticed based on NIDirect health conditions was the total number of results shown was coming from the facet summary block rather than the views header + total rows. That way it lowers the total as you select more facets.

Not holding this up anyway... we can adjust it later if you need this deployed sooner rather than later.

neilblair:
> Looks good, thanks. One thing I noticed based on NIDirect health conditions was the total number of results shown was coming from the facet summary block rather than the views header + total rows. That way it lowers the total as you select more facets.
> 
> Not holding this up anyway... we can adjust it later if you need this deployed sooner rather than later.

Testing locally, the @total rows seems to update correctly when facets are selected.  But if that turns out to be incorrect yes will revisit.

PR #226, opened 2022-11-02
DEPT-232: fix WSOD when selecting facets

neilblair:
Also fixes https://digitaldevelopment.atlassian.net/browse/DEPT-233

The facets_pretty_paths module is breaking facets. Tried updating to the latest patch on https://www.drupal.org/project/facets_pretty_paths/issues/3254600 - but no joy.

Disabling facets pretty paths url processor and using the query string processor instead resolves the problem.  The facets pretty paths url processor can be re-enabled when https://www.drupal.org/project/facets_pretty_paths/issues/3254600 is resolved.


johangant (approved):
Nice one. We'll need to monitor the progress of that issue referenced and incorporate any future patches or releases as they become available.

PR #231, opened 2022-11-04
DEPT-230 ~ Call paths to drush

omahm:


johangant (approved):
Drush path should be already set on the environment but if this works, then let's use it :+1:

PR #236, opened 2022-11-08
Dept 227 fix WSOD on some nodes' display

johangant:
Old values in the db clash with D9 token replacement for embedded media entities, resulting in a very misleading db collation error + WSOD.

Belts 'n' braces approach here which seeks to both update the db table values on deployment / db update, as well as after future migration runs.


omahm (approved):
Nice one

PR #246, opened 2022-11-11
Reconcile two clashing d.o. patches into single patch file

johangant:
Drupal.org issues that are related, but clash a little:

* https://www.drupal.org/node/2952005
* https://www.drupal.org/node/3254600

Fixes facet results count, reset/remove link and compatibility with contextual args on views.


omahm (approved):

johangant:
NB: there are still some issues with combining facets and the paths they apply. Merging this as it solves a few problems in the meantime.

PR #247, opened 2022-11-11
Dept-215

neilblair:
Add sort links to publications search so user can sort results by publication date or by relevance.


omahm (approved):
LGTM, nice one.

PR #262, opened 2022-11-18
Dept-259 press releases search

neilblair:
Add facet summary, fix page title block, add sorting links and css classes for theming

### DIFF
--- a/config/sync/views.view.press_release_search.yml
+++ b/config/sync/views.view.press_release_search.yml
@@ -7,6 +7,7 @@ dependencies:
     - group.type.department_site
     - search_api.index.press_releases
   module:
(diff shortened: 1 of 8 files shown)


config/sync/views.view.press_release_search.yml
```diff
@@ -408,21 +430,25 @@ display:
     display_plugin: page
     position: 1
     display_options:
+      cache:
+        type: none
```
johangant, line 434: Think we usually use search API cache tags for views, unless there's an overriding reason to set no cache on the output.

johangant (approved):
LGTM, just the one point on cache control.

PR #263, opened 2022-11-21
Set PHP memory limits for lando and platform.sh

neilblair:
Is 512M enough for PHP on lando?


johangant:
> Is 512M enough for PHP on lando?

Should be! We've got a fraction of that on the dev cluster so this seems ample right now

johangant (approved):

PR #264, opened 2022-11-21
DEPT-253 Press releases group metadata in the listing

johangant:


web/modules/custom/dept_node/dept_node.module
```diff
@@ -149,11 +150,40 @@ function dept_node_theme($existing, $type, $theme, $path) {
...
   }
 
+  if ($view_mode === 'search_metadata' && $node->bundle() === 'news') {
+    // Add the department metadata on the press-releases listing.
+    if ($routename != 'view.press_release_search.pr_search_page') {
+      return;
+    }
+
+    // Inject the department for a press-release/news items for search.
+    $group_contents = GroupContent::loadByEntity($variables['node']);
```
johangant, line 166: Could have used one of the three field formatters to output this, but actually I just wanted the raw values to pipe into the template and avoid all the associated markup/cruft. There's also - potentially - a bug around some content being tagged with NIGov more than once, hence the exclusion check below just to make sure it matches D7 output.

omahm (approved):
lgtm

PR #271, opened 2022-11-24
Rewrite dept news urls to absolute paths on press releases listing

johangant:
NB: there's enough crossover with RelToAbsUrlsFilter.php to make me think we could do with a service to handle this kind of URL adjustment. The precise return type/end use would still be down to the module/class making use of it, though.


neilblair (approved):
LGTM

PR #272, opened 2022-11-25
Dept 266

omahm:
Bunch of fixes for migrations and a WSOD when config split isn't set


johangant (approved):
Nice one - much more readable migrations list, too!

PR #277, opened 2022-11-30
Dept 267 service object to encapsulate process to convert relative urls to absolute ones

johangant:
Ensures that the source group for a piece of content is linked to from, say, node body (via text filter) or via preprocessed search result row (eg: on press releases when viewed on parent dept; NIGov).


omahm (approved):
Nice one

PR #292, opened 2022-12-09
DEPT-250 Disable Drupal 7 node link block

ooneill547:


johangant (approved):
No problem for me. We can always re-enable or add to a config split profile, if needed

PR #297, opened 2022-12-21
DEPT-256 custom sorted entity queues behind (sub)topics lists

johangant:
Bit of a whopper of a PR which has meandered from being based on entityqueues, back to draggableviews to house the structures of the subtopics lists (as they appear on topic pages + articles under each subtopic) and as a second step - how the topic pages per dept will be ordered.

Currently still relying on entityqueue to provide the topics, but this will be pulled out soon in favour of draggable views based ordering.

Still to follow: replacing any dependencies of groups with domain_access.

This PR mostly intended to be released now to ease any transition across with the group/domain changes scheduled to follow.


web/modules/custom/dept_topics/src/Commands/DeptTopicsCommands.php
```diff
@@ -0,0 +1,301 @@
...
+
+  /**
+   * Builds the top-level queue of items listed under /topics.
+   */
+  public function regenerateTopicsQueue() {
+    $d7topicsByDept = $this->getD7TopicsByDept();
+
+    if (!empty($d7topicsByDept)) {
+      foreach ($d7topicsByDept as $d7_domain_id => $queue_item) {
+
```
omahm, line 71: What's this empty foreach for?

omahm (changes requested):
Looks good, just one issue with the foreach loop

omahm (approved):
LGTM, nice one

PR #314, opened 2023-01-23
Dept 256  subtopics with articles and other subtopics

johangant:
Now that content under subtopics can also reference articles and other subtopics we had to reshuffle how those lists were created. Where there isn't draggableviews data to determine the precise order of things to be returned we fall back to a query that unions articles and subtopics referencing a subtopic and use the results to generate the same result set for the lists we show.

### DIFF
--- a/web/modules/custom/dept_node/src/ContentTopics.php
+++ b/web/modules/custom/dept_node/src/ContentTopics.php
@@ -166,4 +166,53 @@ class ContentTopics {
     return $render;
   }
 
+  /**
+   * @param int|\Drupal\node\NodeInterface $node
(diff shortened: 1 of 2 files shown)


neilblair (approved):
Just wondering whether the UNION can be done using the db api? If not, or performs badly, looks good to me.

johangant:
> Just wondering whether the UNION can be done using the db api? If not, or performs badly, looks good to me.

You can create a dynamic query with a union, but it's really rather messy and seems more of an afterthought in the DB API. It's more performant to have a single static query with placeholders using the DB API than the dynamic equivalent or, worse, two separate queries + munge together with PHP in a preprocess function.

PR #322, opened 2023-01-31
Dept services refactor

johangant:
A few things I wanted to sort out:

- Make the services clearer about what they do; topics or subtopics.
- Rely on preprocess hooks to handle render elements and let the services return data where possible.
- Pull together the entity queries and ad hoc SQL into service functions to make things more manageable and consistent.


web/modules/custom/dept_topics/src/ContentSubTopics.php
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Constructor for ContentTopic class.
```
omahm, line 24: Should be 'ContentSubTopic'

omahm (changes requested):
Nice refactor. Just one tiny copy and pasta change.

web/modules/custom/dept_topics/src/ContentSubTopics.php
johangant: Good spot, will get that amended!

omahm (approved):

PR #333, opened 2023-02-09
Draggable views data sync and adjusted topics config

johangant:
Resets some missing display options for Topics, and adjusts the drush command used to pull draggable views data from D7.


web/modules/custom/dept_topics/src/Commands/DeptTopicsCommands.php
```diff
@@ -0,0 +1,351 @@
...
+  /**
+   * Barebones function to return a dept id based on D7 domain id.
+   *
+   * @param int $d7_id
+   *   The D7 domain id.
+   *
+   * @return string
+   *   The D9 dept id.
+   */
+  private function domainD7ToD9(int $d7_id) {
```
johangant, line 84: Couldn't find something open/service related that did the same thing as this so popped it in here (for now) - should be easy to extract into a service (new or existing) if it's useful in multiple contexts

web/modules/custom/dept_topics/src/Commands/DeptTopicsCommands.php
```diff
@@ -0,0 +1,351 @@
...
+        $d9_topic_id = $d9_topic_lookup['nid'];
+
+        $row = [
+          'view_name' => 'content_stacks',
+          'view_display' => 'dept_topics',
+          'args' => '["' . $d9_dept_id . '"]',
+          'entity_id' => $d9_topic_lookup['nid'],
+          'weight' => $topic_data['weight'],
+        ];
+        \Drupal::database()->insert('draggableviews_structure')
```
omahm, line 122: I'd look at injecting the database connection as a service

web/modules/custom/dept_topics/src/Commands/DeptTopicsCommands.php
```diff
@@ -0,0 +1,351 @@
...
+        $d9_subtopic_id = $lookup_data['nid'] ?? 0;
+
+        if (empty($d9_subtopic_id)) {
+          $this->io()
+            ->writeln("No D9 subtopic node found for D7 id " . $d7_subtopic_id);
+          continue;
+        }
+
+        // Is there a draggable views entry?
+        $existing_dv_data = \Drupal::database()
```
omahm, line 171: Inject this and the call further down as a service

omahm (changes requested):
Looks good overall but I'd maybe inject the db conn instead of calling Drupal::database each time in those for loops.

omahm (approved):
LGTM 👍

PR #342, opened 2023-02-14
Apply patch to admin toolbar to fix too-long submenus

neilblair:


neilblair:
Just realised patch isn't actually doing anything ...

omahm (approved):

PR #345, opened 2023-02-16
DEPT-343 "We work with these bodies" (agency links) block

neilblair:
plus admin links to draggable links views plus other tweaks

### DIFF
--- a/config/sync/block.block.department_details_contact_us.yml
+++ b/config/sync/block.block.department_details_contact_us.yml
@@ -1,25 +1,33 @@
-uuid: c36a96d3-6ad0-4b4d-81ed-8fccb8ae30c9
+uuid: e2c1e57e-7a1e-4155-b346-0def4648579e
 langcode: en
 status: true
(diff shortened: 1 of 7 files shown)


config/sync/block.block.department_details_contact_us.yml
```diff
@@ -1,25 +1,33 @@
...
 region: content
-weight: -15
+weight: -13
 provider: null
 plugin: dept_core_departmental_details
 settings:
   id: dept_core_departmental_details
-  label: 'Contact us'
-  label_display: visible
+  label: 'Department details - contact us'
```
ooneill547, line 19: Did you mean for this to revert?

ooneill547 (approved):

config/sync/block.block.department_details_contact_us.yml
neilblair: Yep - think so ... not sure now!  Will change later if needs be.

PR #358, opened 2023-02-24
DEPT 328 structured address fields

johangant:
* Moves the legacy address field out of sight on the editorial forms but retained for reference if needed in future.
* Replaces the text fields for address parts with a proper address component.
* Integration with geolocation module for accurate and simple geocoding.
* Map widget for address disabled on heritage site, users encouraged to populate the address component fields. Clicking a point on the map geocodes the nearest known address and populates the address component (two-way sync)


omahm (approved):
Nice one. Far better to have that data stored in a proper format

PR #366, opened 2023-03-01
Introduce dept id and name as preprocess variables for use in templates

johangant:
Some important points about Twig allowed method names config to be aware of, but not of any direct consequence here as I've opted to supply the commonly used dept values as preprocess variable values rather than Department objects


johangant:
Basic usage within Twig templates (page and node templates - corresponds to preprocess_page and preprocess_node hooks) like so:

`{% if dept.id == 'finance' then %}{{ dept.name }}{% endif %}`

ooneill547 (approved):

PR #367, opened 2023-03-01
Set node and content stack cache tags per subtopic list

johangant:
Ensures that the elements are correctly expired when:

- Content is added to the subtopic list.
- Content in a stack is moved in the view config.
- Content is removed from the stack
- Content is changed (eg: title)


omahm (approved):
Looks good, quite the work around!

PR #385, opened 2023-03-09
Remove devel from prod config

johangant:


ooneill547:
I think you need to include devel generate also

omahm (approved):

johangant:
Added devel_generate to the list of removals. Config imports cleanly so will merge now.

PR #387, opened 2023-03-10
Dept 382

omahm:


johangant (approved):
Nice! This might be something we can use instead of the preprocess hooks to pipe the department object towards the template layer, if we need to do much more complex work with that.

PR #395, opened 2023-03-14
Dept 385

neilblair:
News display tweaks.

### DIFF
--- a/config/sync/system.performance.yml
+++ b/config/sync/system.performance.yml
@@ -2,9 +2,9 @@ _core:
   default_config_hash: b2cssrj-lOmATIbdehfCqfCFgVR0qCdxxWhwqa2KBVQ
 cache:
   page:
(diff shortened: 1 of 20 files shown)


config/sync/system.performance.yml
```diff
@@ -2,9 +2,9 @@ _core:
   default_config_hash: b2cssrj-lOmATIbdehfCqfCFgVR0qCdxxWhwqa2KBVQ
 cache:
   page:
-    max_age: 31536000
+    max_age: 0
```
johangant: Do we not want a long TTL for cache here? NIDirect has > 0 at https://github.com/dof-dss/nidirect-drupal/blob/development/config/sync/system.performance.yml

johangant (approved):
LGTM (ignore last comment about cache TTL - it's out of date)

PR #396, opened 2023-03-14
Dept 333

omahm:


johangant (approved):
LGTM - a type hint on the dept entity might be needed to instruct phpstan what object it's supposed to have at that point

PR #413, opened 2023-03-27
Dept 404

omahm:
Duplicating the sync/ domains into local to let config filter load those hostnames along with the 2 other split environments.


johangant (approved):
LGTM

PR #428, opened 2023-04-05
Site search tweaks

neilblair:
Fixes issue where search input is empty after performing a search.


config/sync/block.block.footer.yml
```diff
@@ -11,7 +11,7 @@ dependencies:
 id: footer
 theme: nicsdru_dept_theme
 region: footer
-weight: -25
+weight: -23
```
ooneill547, line 14: I don't think this should be changed

ooneill547 (approved):
Just need to remove the weight change in the footer

config/sync/block.block.footer.yml
omahm: Is this to pop the footer links (copyright etc) above the department links?

neilblair: Oopsy fixed ☝🏻

PR #431, opened 2023-04-06
Dept 398 - search autocomplete block + related config

johangant:
- Same approach used as per NIDirect
- Autocomplete config was awkward with preprocessing as we want to avoid loading node entities to then append the dept label on NIGOV.
- Adding the domain source value to the Solr document wasn't enough as the module's theme element prefers an entity display mode to render the value.
- Accordingly, a new view mode for search autocomplete is used to append the domain source field. We preprocess this on non-NIGOV domains to remove it from the render array and there's a template for both autocomplete result item and new view mode for nodes which is lightweight and flexible to use.


web/modules/custom/dept_node/dept_node.module
```diff
@@ -127,6 +127,12 @@ function dept_node_preprocess_node(array &$variables) {
   // Add the topic banner to the output, if relevant.
   dept_node_topic_banner($variables);
 
+  // Exit early if it's autocomplete - see dept_search.module
+  // preprocess_node for handling of autocomplete results.
+  if ($view_mode === 'search_autocomplete') {
```
omahm, line 132: Any reason not to have this before the dept_node_topic_bannner call?

omahm (commented):
Looks good, just one query.

web/modules/custom/dept_node/dept_node.module
johangant: Agreed, this would be better. Good spot! I'll move this up

PR #433, opened 2023-04-12
Fallback entity query if topic doesn't have subtopics defined in draggable views

johangant:
We had a similar edge case in subtopic listings where some articles weren't 100% accurate when syncing from D7. TLDR: we'd bring across what we reasonably can, then have site owners adjust/set this before launch after which the issue resolves itself and concentrates technical effort on higher value tasks.


omahm (approved):
lgtm

PR #437, opened 2023-04-17
Add text filter to convert absolute urls on the same dept site to relative ones

johangant:
Differs from the RelToAbsUrl filter as this goes the other direction + is used for converting links to absolute values on top-level sites like NIGOV.


web/modules/custom/dept_core/modules/dept_content_processors/src/Plugin/Filter/AbsToRelUrlsFilter.php
```diff
@@ -0,0 +1,133 @@
...
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\dept_core\DepartmentManager $department_manager
+   *   The department manager object.
+   *
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, DepartmentManager $department_manager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
```
omahm, line 48: I'd look at using the $instance pattern which doesn't override the base constructor.

web/modules/custom/dept_core/modules/dept_content_processors/src/Plugin/Filter/AbsToRelUrlsFilter.php
```diff
@@ -0,0 +1,133 @@
...
+        if (str_starts_with($href, 'http')) {
+          // Look at the hostname, if it matches our current domain
+          // then remove the hostname from the attribute and update the
+          // dom element.
+          $url_portions = parse_url($href);
+          $host = $url_portions['host'];
+
+          if (!empty($url_portions['path']) && $this->hostnameMatchesKnownDepartment($host, $dept->id())) {
+            $new_href = $url_portions['path'];
+            if (!empty($url_portions['query'])) {
```
omahm, line 91: I'd also check for the hashbang

omahm (changes requested):
Looks good, just a couple of comments

web/modules/custom/dept_core/modules/dept_content_processors/src/Plugin/Filter/AbsToRelUrlsFilter.php
johangant: Having another attempt at this today but it's functional with just using the dept manager service, so would see the context injection as an enhancement.

omahm (approved):
LGTM

PR #449, opened 2023-04-25
DEPT-285 secure publications

johangant:
Intended to:

- Restrict access to publication nodes that have a populated secure files field (designates as secure publication) value. Full site(s) admins can view the content by default but general editors should not be able to view unpublished revisions.
- Users with the stats author role can create new publication nodes but access to unpublished revisions should be restricted to them, stats supervisor roles and high level admins (expected to be webdev staff).
- Functional tests are left as they are. They're partially working but depends on being able to cleanly install from current config and not from a separate install profile.


omahm (approved):
Nice one Johan

PR #451, opened 2023-05-03
Dept 299

neilblair:
Initial setup of layout builder on subtopic pages.


web/modules/custom/dept_landing_pages/dept_landing_pages.info.yml
```diff
@@ -0,0 +1,8 @@
+name: 'Dept Landing Pages'
```
omahm, line 1: I'd rename this to 'Departmental sites: Landing pages' to keep it consistent with the other modules

omahm (changes requested):
Has the campaign importer modules been tested to see if it works? I remember writing that a long time ago for D8 and I couldn't guarantee it's up to date for D9

omahm (approved):
LG

web/modules/custom/dept_topics/src/Plugin/Block/TopicContentBlock.php
```diff
@@ -0,0 +1,143 @@
...
+        '#title' => $node->getTitle(),
+        '#attributes' => [
+          'class' => ['subtopics'],
+        ],
+        'items' => [],
+      ];
+
+      foreach ($subtopics as $subtopic_id => $subtopic) {
+
+        $subtopic_content = $this->subtopics->getSubtopicContent($subtopic_id);
```
omahm, line 104: Just a heads up that this method probably won't exist in the new topic setup

PR #454, opened 2023-05-15
DEPT-420 improve variable PSH hostname handling

johangant:
Bit of a workaround but this seems to work by detecting the PSH env vars and then replacing the hostname found from the config YML files when the Department entity hostname is generated.

It won't kick in on PSH production environments, nor local development as that has its own config split profile.


omahm (approved):
Nice one 👍

PR #455, opened 2023-05-18
DEPT-450 

johangant:


web/modules/custom/dept_publications/dept_publications.module
```diff
@@ -122,3 +123,27 @@ function dept_publications_entity_access(EntityInterface $entity, $operation, Ac
     }
   }
 }
+
+/**
+ * Implements hook_entity_operation_alter().
+ */
+function dept_publications_entity_operation_alter(array &$operations, EntityInterface $entity) {
```
johangant, line 130: Belts 'n' braces here. We're aiming to cover someone trying to view the node directly, as well as any operations that might be defined for it, as well as mediate in the origins workflow access control for changing to - specifically - published moderation state.

web/modules/custom/dept_publications/src/Routing/SecurePublicationsRouteSubscriber.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+class SecurePublicationsRouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+    if ($route = $collection->get('origins_workflow.moderation_state_controller_change_state')) {
+      $route->setRequirements([
+        '_custom_access' => '\Drupal\dept_publications\Controller\SecurePublicationsPermissionController::accessCheckModerationStateChange',
```
johangant, line 19: YOINK!

We step in and hijack the existing controller definition. Here we can then evaluate our own requirements and, if all else fails, proxy on the request to the existing origins workflow controller and let it continue to do its thing.

neilblair (approved):
LGTM

PR #471, opened 2023-06-06
Dept 469

omahm:
Flag option to hide a Topic/Subtopic is retained for use in the help topics system.


neilblair (changes requested):
Think we may need to keep field_global_topics ... see thread in slack

neilblair (approved):

PR #476, opened 2023-06-08
Dept 463

omahm:


web/modules/custom/dept_node/dept_node.module
```diff
@@ -294,3 +295,22 @@ function dept_node_entity_type_alter(&$entity_types) {
...
   }
 }
+
+/**
+ * Implements hook_pathauto_alias_alter().
+ */
+function dept_node_pathauto_alias_alter(&$alias, array &$context) {
+  $dbConn = \Drupal::database();
+
+  $results = $dbConn->query("SELECT pa.path, pa.alias FROM path_alias pa WHERE pa.alias = :alias", [':alias' => $alias])->fetchCol();
```
johangant, line 305: Curly braces on the table name recommended, so should be `{path_alias} pa`

web/modules/custom/dept_node/src/Form/UpdatePathAliasForm.php
```diff
@@ -0,0 +1,163 @@
...
+
+    if ($results) {
+      $form_state->setErrorByName('new_alias', 'Alias already in use, please try another');
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
```
johangant, line 160: Does this need to do anything? Ie: would it need `parent::submitForm()`? Or just needed to satisfy the interface definition?

johangant (approved):
Looks good. Very minor thing about table prefix wrappers and a question about the submit function that looks empty

web/modules/custom/dept_node/dept_node.module
omahm: good spot !

PR #483, opened 2023-06-14
Dept 442

omahm:


johangant (approved):
LGTM. Good catch restricting the message to web requests so drush doesn't spam the console

PR #486, opened 2023-06-14
DEPT-298 Language handling for Irish news content

johangant:


omahm (approved):
LGTM... and that return comment 🤣

PR #495, opened 2023-06-26
DEPT-502 topic tags not displaying

neilblair:
Update dept_topics.module so that topics and subtopics are listed for nodes using field_site_topics and field_site_subtopics - e.g. publications, consultations and news


omahm (approved):
LG

PR #500, opened 2023-06-27
DEPT-481 only insert new topic content if it's unique

johangant:
Needs the `node__field_topic_content` table truncated after deployment to clear out the existing duplicates.


omahm (approved):
LG

PR #503, opened 2023-07-03
Domain enabled media entities

johangant:
- Adds a domain source + access field for all media entity types.
- Some preprocessing on media library views displays to introduce a better filter; pre-set/hidden value for non-admin users.
- Migrate process plugin for media content when importing so that it sets the correct domain value based on file usage in D7

NB: After deployment set all source_row_status values to 1 for each media type. This will then update each item from source during the overnight import tasks. May be prudent to do one type per night to avoid excessively long processing times.


web/modules/custom/dept_core/dept_core.module
```diff
@@ -256,3 +258,38 @@ function dept_core_department_links(): array {
...
+      if (!\Drupal::currentUser()->hasPermission('administer domains')) {
+        /** @var \Drupal\dept_core\Entity\Department $dept */
+        $dept = $dept_manager->getCurrentDepartment();
+        $form['dept']['#value'] = $dept->id();
+        hide($form['dept']);
+        return;
+      }
+
+      // Change to select list of known depts.
+      $depts = \Drupal::service('department.manager')->getAllDepartments();
```
omahm, line 281: You're loading this service twice, I'd replace it with $dept_manager->getAllDepartments();

omahm (changes requested):
Nice one, good job 👍  
Just 1 wee minor request.

web/modules/custom/dept_core/dept_core.module
johangant: Good spot, will pop that in now!

johangant:
Service adjustment added, will merge now

PR #516, opened 2023-07-21
DEPT-501 change summary field from formatted to plain text

johangant:
Quite a few knock-on changes here:

- Update hook to convert the field and values (cleaner than manual approach to remove/replace and repopulate from migration)
- Field storage/instance and form display config
- Migration config changes to reflect the different field type (shoehorned in the editors notes field as this wasn't set properly [separate issue fixed along the way])


omahm (approved):
LGTM

PR #517, opened 2023-07-25
CI build constraints

johangant:
Initial PoC using a local constraints file against the composer.lock contents in the repo


johangant:
Looked at adding jq + dependent packages via curl downloads from packages.ubuntu.com but it looks far more practical to stick with installation via apt / synaptic to ensure all current and future dependencies are included.

Next steps outside of this project would be to consider how to manage and incorporate the information from constraints.txt

omahm (approved):
Nice one

PR #521, opened 2023-08-02
Dept 513

omahm:


johangant (approved):
LGTM. Like the use of the event to avoid any accidental clashes with more general entity hooks. Can't spot any obvious issues in there so I think review on edge when it merges overnight and see how it goes :+1:

PR #522, opened 2023-08-07
Adjust topic picker widget to a permission for finer grained control

johangant:
Prior problem was that stats author (or similar future roles less than full author) roles could not use the topic picker without being given node permissions for topics/subtopics too which wasn't in the brief for those restricted roles.


omahm (approved):
Funny, I had considered a custom permission for this.

ooneill547 (approved):

PR #532, opened 2023-08-24
Add event subscriber to listen for/adjust site topics field values on Solr docs

johangant:
Once applied, single topic selections against field_site_topics use the topic manager service to look up the tree and also add parent topics. This ensures that the publications/consultations searches surface content correctly.

Eg:

**Before:**

image

**After:**

image


johangant:
NB: I've kept the set values as the underlying topic ids. There's also a text field: `tm_X3b_en_site_topics_title` which we might need to extend this to update too, depending on how it might be used later on. That shouldn't be too awkward to fold in later if needed.

neilblair (approved):
LGTM

PR #543, opened 2023-09-08
DEPT-534 topics modal window tweaks

johangant:


johangant:
Works in combination with #541

omahm (approved):
👍

PR #586, opened 2023-10-18
DEPT-573 facet topics hierarchy preprocessing

johangant:
Topics are not taxonomy based, rather they're nodes that reference other topic/subtopic nodes.

This means we can't use the out of the box hierarchy support for taxonomy based facets. Initially we get a big flat list of all the topics which isn't that helpful. This PR fudges this a little by removing unwanted facet items depending on what, if anything, has been selected. Applies to publications and consultations search.


omahm (approved):
Nice, LGTM

PR #593, opened 2023-11-01
Move related content block adjustments into views hook from theme preprocess

johangant:


web/themes/custom/nicsdru_dept_theme/inc/preprocess.inc
```diff
@@ -192,33 +188,6 @@ function nicsdru_dept_theme_preprocess_page(array &$variables) {
             $variables['page']['sidebar_second'][] = Helpers::blockContent('featurednews');
             $variables['page']['sidebar_second'][] = Helpers::blockContent('views_block__news_latest_news_block');
           }
-
```
johangant, line 195: Pulled this out as the core block functionality appears to now put the correct things into place where it's configure to appear. The exception is, still, the preview path where some creativity is needed to derive the node and pipe in the correct contextual argument value

johangant: Opted for this approach too. We can perhaps add a representative block in the sidebar in future if it proves useful

PR #601, opened 2023-11-08
Ckeditor5 upgrade

DuttonMa:
- Uninstall origins CKEditor 4 module
- Install supporting modules for CKEditor 5
- Convert text formats 'basic html', 'full html' and 'layout builder html' to CKEditor 5


omahm (approved):
LG

PR #621, opened 2023-11-30
DEPT-611

johangant:
Gnarly mixture of config adjustment and something funny happening to the value of <front>. Likely triggered by domain_source module but for the template I've passed in a dept manager URL value which is less likely to be tampered with in future, and fixes the issue.


omahm (approved):
lgtm

PR #626, opened 2023-12-11
DEPT-611

johangant:
Something of a cludge to rewrite the URL in the main menu. Very opaque as to where, precisely, domain/domain_source adjusts the routing based on the canonical domain set for content so I'd consider this a workaround than a preferred fix.

This ticket also adds an extra default cache context based on the domain hostname, so a variant of a render element will usually be produced for each domain rather than one domain's HTML being re-used across all domains.


omahm (approved):
nice work around.

PR #629, opened 2023-12-15
DEPT-631

johangant:
- Exclude publication nodes with secure files attachments from:
  - Main content listing
  - Media assets from media library 
- Default publication node form to secure publication fields if the user has the permission to do so


omahm (approved):
Nice one, I bet that was fun.

PR #636, opened 2023-12-20
DEPT-638 PoC for ignoring certain domain ids during migrations

johangant:
Had a couple of approaches with this:

- Adding extract/flatten ahead of the sub_process plugin for handling domain_access or source. Turns out that mangles the final value coming through, so:
- Introduce a new process item which doesn't map to a destination field, but is evaluated. That way we can look at the canonical source of an item of content and then flatten the entire array structure and look for known values to match (live dept site ids) and skip the row if a match is found.


omahm (approved):
That's a really neat solution!

PR #643, opened 2024-01-03
DEPT-631 secure pubs tweaks

johangant:
Only simplify the form for secure publications if the logged in user has the permission to create new secure publications and is not already a top level administrator (automatically assigned all new permissions)


omahm (approved):
LGTM

PR #648, opened 2024-01-05
Migration taxo issues

johangant:
- Replace needless migrate lookups with fixed values
- turn off content lock modules at start of nightly migration (re-enabled at the end by config import)


omahm (approved):
LGTM

PR #651, opened 2024-01-09
DEPT 631

johangant:
- New service object to oversee private file access control
- Filter main content listing to weed out secure publications if permissions not in effect
- Add extra condition to media library and remove if user can see secure publication assets
- Ignore inspection of where condition; confuses phpstan
- Only simplify the publications form for secure pubs if user is not an administrator
- Exclude secure publications from moderation view displays
- Make secure publication field mandatory for non-admin users


omahm (approved):
LGTM

PR #652, opened 2024-01-10
Fix array error with WYSIWYG migration filter

omahm:


johangant (approved):
Ah, nice one. Cleaner log output now :)

PR #662, opened 2024-01-23
DEPT-644 ~ Improve node edit pages

ooneill547:


omahm (approved):
👍

johangant (commented):
Looks good. If I was being fussy I'd say remove the commented lines of SCSS, but that's being really fussy :smile:

PR #667, opened 2024-02-02
DEPT-652 Fix UAL photo migration mappings

johangant:


johangant:
Once deployed need to run migrate:import node_ual --force --update to refresh existing content.

omahm (approved):

PR #691, opened 2024-02-28
Dept 674

omahm:
Remove geolocation fields and replace with a text field using the simple gmaps formatter. Update migrations to concat lat/log into coords for textfield.


johangant (approved):
LGTM. Worth keeping in the existing geolocation module for now until this has settled, then we can uninstall and remove the old module without the front falling off :)

omahm:
> LGTM. Worth keeping in the existing geolocation module for now until this has settled, then we can uninstall and remove the old module without the front falling off :)

Exactly why I've left it in as I can imagine cim screaming in panic that the module has gone!

PR #703, opened 2024-03-11
Dept 686

omahm:


johangant (approved):
Nice, let's see how it works in prod :+1:

PR #709, opened 2024-03-14
Dept 689

omahm:
Prevent book pages from appearing in topic child content.


johangant (approved):
Nice catch here. Looks great

PR #723, opened 2024-03-26
DEPT-653 ~ Change code to stop clientside validation on node edit forms

ooneill547:
I changed the code to a more basic version that seems to work and also added in media forms as they were having the same problems.


johangant (approved):
Might be worth moving this into dept_core.module if it's moving beyond nodes alone. This module's always caused a few headaches in unexpected places

PR #742, opened 2024-04-08
DEPT-671 Table wide replace pattern

johangant:
One hit fix for wildcarded hostname pattern in sitemap index column. Reduces processing overhead; previously loading each dept in sequence with update sql command for each.


omahm (approved):
lgtm

PR #751, opened 2024-04-10
Dept 714

omahm:
- Filter allows subdomains of the department domain.
- Updated filter to allow testing.
- Removed test suites as they fail when indexing tests because some contrib module tests don't conform to the parent base test method signature.


ooneill547 (approved):

johangant (approved):
Looks good, nice to see some extra testing coverage go in 👍

PR #753, opened 2024-04-12
Layout builder CSS overrides

johangant:
Proof of concept. Sizes may need adjusting and could be useful to move, if successful, into a new module... eg: dept_layout_builder


johangant:
@ooneill547 I think this should now let you experiment with SCSS mixins and adjustments. Compilation will output a CSS file in the usual theme location, and there's a theme library added which is then dovetailed in with a form_alter hook in the theme alter.inc file on the layout builder forms. Hopefully this bridges more of the gap between your want for more SCSS integration and control. NB: the precise values in there will need some adjustment but you can be confident that particular CSS file will only be added to the render arrays we tell it to in the form_alter hook.

PR #757, opened 2024-04-15
Dept 721

omahm:
- Limit topic selection for subtopics.
- Fix treeview ID for child content.
- Update modal with close ajax command.


johangant (approved):
LGTM

PR #772, opened 2024-04-26
Disable link rewrite/redirect for node canonical paths

johangant:
Turn this off and it means a node shared to a domain can be seen on that domain, instead of redirecting to the hostname defined from the domain source field value.


omahm (approved):
Sweeet.

PR #776, opened 2024-04-30
DEPT-748

neilblair:
Where a subtopic page has content in the body field:
A “More in this topic” heading is shown above the hyphenated bulleted list of child content links that follows the body content.

Where a subtopic page does not have content in the body field:
The hyphenated bulleted list of child content links is shown below the title and summary without any heading.

For topic or subtopic pages:
- the heading for the section containing recent news, publications or consultations is “Most recent …”
- the “Most recent …” heading is not shown when “More in this topic” is displayed.


omahm (approved):
LG

PR #796, opened 2024-05-16
Process plugin for body content for improved visibility and logging

johangant:
Config for node_article migration


omahm (approved):
Nice one. Yeah the body_field_process makes sense now as we're gradually added more and more processing to the body field contents.

PR #800, opened 2024-05-20
Patch filelog module to fix source of circular reference bug

johangant:
See https://www.drupal.org/project/filelog/issues/3416342


omahm (approved):
nice one

PR #809, opened 2024-05-30
DEPT-780 ~ Add book navigation cache setting

ooneill547:
I'm not sure if this is the correct approach, it is what has been implemented in unity


johangant (approved):
This might be a smidge heavy handed if it forces any page it appears on to become uncachable. We can always adjust cache policy on that render element if we notice any adverse effects later, though

PR #812, opened 2024-06-06
DEPT-796 adjust [[fid]] tokens and adjacent html tag ordering

johangant:
Enclosing [[fid]] in <p> (for example) leads to [[fid]] being replaced with <drupal-media> in D10, which when rendered will obscure or break the content showing alongside that token. Moving the [[fid]] token outside a parent element results in content showing fully with no other display issues.


ooneill547 (approved):
This looks very complicated. Well done. Hopefully it resolves the issues

PR #825, opened 2024-06-21
DEPT-780 refactor and adjust cache busting when book outline changes or content is removed

johangant:
Book module expires 'bid:123' style keys, but not the related node tag meaning changes aren't reflected right away when adjusting a book outline.

Had to shoehorn in a small patch to core to get this to work as post-submit handlers didn't seem to work. Doubled up with suitable, related cache tags added to the book nav block and expirations on the book outline removal form, too.


omahm (approved):
Looks good. I've a bunch of code to fix another Book issue and I'll move that into this module.

PR #849, opened 2024-07-19
Attach custom JS to toolbar render element to fix hostname discrepancies

johangant:
Clientside workaround that checks the paths of any absolute links in the toolbar, except the 'Sites' menu. If any absolute links are found then they're replaced with values using the current site's hostname.

image

NB: this should be managed by domain/domain_source modules but due to some very opaque issues with the way cache items are generated, keyed and re-used in those modules the default domain sometimes is used for the <front> route path. Not a lot we can do server side to preprocess this owing to the way route paths are cached and altered by domain_source module. Even altering the toolbar render element doesn't seem to work around the issue. This could, potentially, pop up in other cases where something might need an absolute URL to the current department's front page but in those instances I would suggest using a relative path when generating the URL, as and when required.

### DIFF
--- a/web/modules/custom/dept_homepage/dept_homepage.info.yml
+++ b/web/modules/custom/dept_homepage/dept_homepage.info.yml
@@ -1,5 +1,5 @@
 name: 'Department Homepage'
 type: module
 description: 'Functionality relating to the site homepage'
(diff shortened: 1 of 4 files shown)


neilblair (commented):
Belated LGTM!

PR #855, opened 2024-07-24
DEPT-840 Migrate audit report

johangant:
Adds:

- Admin page in D10 to show orphaned content in migration tables, with links to D10 and D7 paths to allow subsequent edit/removal as needed.
- Report by content type (basic links and route param used to split things up), with pager for longer result sets.
- Tells you total amount and when it was last audited.
- Cron handler that fetches uuid data from D7 every 3 hours, populates an internal table in D10 to use as a diff between the required migrate_map_node_$type table.
- Cron handler queries take about 5 seconds to run in total.

image


johangant:
BTW if you run the cron task via the UI, it'll hit the buffers re: memory limits. So it has to run via drush/CLI which has the expanded memory allocation

johangant:
Merging in to make the overnight build!

PR #856, opened 2024-07-29
DEPT-834 ~ Make operations dropdown the same size

ooneill547:


web/modules/custom/dept_core/dept_core.module
```diff
@@ -399,3 +399,11 @@ function dept_core_clientside_validation_should_validate($element, FormStateInte
     return TRUE;
   }
 }
+
+/**
+ * Implements hook_page_attachments().
+ */
+function dept_core_page_attachments(array &$attachments) {
+  // Attach extra custom css for admin menu.
+  $attachments['#attached']['library'][] = 'nicsdru_dept_theme/admin-content';
```
omahm, line 408: I'd use the AdminContext class so this css is only applied to admin routes.
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Routing%21AdminContext.php/function/AdminContext%3A%3AisAdminRoute/10

omahm (approved):

PR #866, opened 2024-08-07
Fill in the remote doc file mimetype value

johangant:


omahm (approved):
Nice one, looks good.

PR #867, opened 2024-08-08
Add publish status filter so topics shown on /topics must also be pub…

johangant:
…lished


omahm:
The published state was removed so that editors could see unpublished topics on the /topics page. The public won't see this as they don't have the 'view unpublished' permission

johangant:
Ah, yes I remember this now. Well spotted - this isn't needed so closing the PR

PR #869, opened 2024-08-09
Dept 843

omahm:
-- Option to delete selected nodes.
-- If selected nodes have media entities, check if they exist elsewhere and if not, delete.
-- Integrate audit process batch call into form.
-- Styling changes.


johangant (approved):
LGTM, nice one!

PR #876, opened 2024-08-22
DEPT-841 replace Lando with DDEV

johangant:
**Some big changes. Key highlights:**

- DDEV installs mkcert which gets around the annoying self-certified warning on Lando
- Mutagen and Rancher provides better performance than Docker Desktop alone.
- DDEV hooks provide lots of flexibility for improving onboarding process.
- DDEV addons such as cron provide easy local extensibility and services.

**One major change:**

- Migrations don't use a separate db service, but DEPT on PSH does. We'll keep it that way but you'll notice the migrate group config patches in the local config split profile. These should import for you on db import success.


omahm:
Looks good but holding off to see if there is the budget for Orbstack.

DuttonMa (approved):
Great stuff !

PR #895, opened 2024-09-11
DEPT-876

omahm:


ooneill547 (approved):

.ddev/config.yaml
```diff
@@ -31,8 +31,8 @@ hooks:
     # Import any local config split profile overrides.
     - exec: "drush cim -y"
 name: dept
-nodejs_version: 20
-php_version: "8.2"
+nodejs_version: 14
```
ooneill547, line 34: Will this break the npm fixes Johan did last week?

omahm: Not sure. I went with the most recent looking changes (the PHP version update)

PR #900, opened 2024-09-17
DEPT-894 ~ Topic summary output

ooneill547:
I'm not sure if this is the best approach or if there is maybe a cleaner way to do this?


omahm (approved):
LGTM 👍

PR #903, opened 2024-09-23
Dept 890

omahm:


web/modules/custom/dept_migrate/src/LookupHelper.php
```diff
@@ -0,0 +1,87 @@
...
+      return new LookupEntry($this->lookupManager->lookupByDestinationUuid([$uuid]));
+    }
+  }
+
+  /**
+   * Lookup result using source (migration) database.
+   *
+   * @return $this
+   */
+  public function source() {
```
johangant, line 72: Handy to have these explicit levers to say which direction the lookup is set to go in :)

web/modules/custom/dept_migrate/src/LookupEntry.php
```diff
@@ -0,0 +1,78 @@
...
+class LookupEntry {
+
+  /**
+   * Lookup data.
+   *
+   * @var false|mixed
+   */
+  protected $item;
+
+  public function __construct(array $lookup_item) {
```
johangant, line 14: Safe to assume this is build for a single lookup but has array set to slot in better with other code? Ie: no implementation to handle multiple items at this time, but could be added another day if really needed?

johangant (approved):
This is great. Makes it a lot clearer about the mechanics of looking up content as well as a cleaner implementation on what we had before

web/modules/custom/dept_migrate/src/LookupEntry.php
omahm: I'm just passing the raw results from the lookup service which always returns an array. I had thought about creating a custom collection type but for the time being we don't really need it.

PR #908, opened 2024-09-24
DEPT-897 ~ Hide related region on landing pages

ooneill547:


web/themes/custom/nicsdru_dept_theme/inc/preprocess.inc
```diff
@@ -71,6 +71,13 @@ function nicsdru_dept_theme_preprocess_page(array &$variables) {
           $variables['main_content_attr']->addClass(LayoutClasses::MAIN_LIMITED);
           break;
       }
+
+      // If node has custom layout (landing pages), then hide the related region.
+      $variables['landing_page'] = $node->layout_page ? TRUE : FALSE;
```
omahm, line 76: Not sure we need this added to the vars, could use the null safe operator within the if statement?

ooneill547: True, I'll fix it now

PR #915, opened 2024-09-26
DEPT-907 ~ Override a11y_form_helpers fieldset template

ooneill547:
This will allow descriptions to appear just under the legend in fieldsets on the node edit forms


omahm (approved):
LGTM

PR #930, opened 2024-10-03
Dept 922

johangant:
MIGRATE_IGNORE_SITES needs to be carefully set: no spaces. 

MIGRATE_IGNORE_SITES=finance or MIGRATE_IGNORE_SITES=finance,communities for multiple depts.

Tested locally with users and nodes. Adding depts increases the ignored count each time, which is what we want. Also ignores if you pass the --update flag into migrate:import, which helps guarantee live data.


omahm (approved):
LGTM

PR #931, opened 2024-10-03
Dept 926

omahm:


johangant (approved):
LGTM. Does the submodule for PSH tools need anything else to work or will it clone with each git pull on the repo?

omahm:
> LGTM. Does the submodule for PSH tools need anything else to work or will it clone with each git pull on the repo?

It's just a clone of the original repo.

PR #940, opened 2024-10-08
Dept 908

omahm:
- Setup sitemaps for each content type and department but disable non-live department sitemap creation.
- Hook to insert each department homepage into the sitemap because we're using a controller for that page.
- Fixed 'Passing null to parameter #1' by not having Domain Simple Sitemap not generate the homepage.
- Removed patch for above issue as we don't need it because it doesn't work in our case which is why we we are generating our own homepage links.


johangant (approved):
That's cool, a lot cleaner than the previous workaround!

PR #941, opened 2024-10-08
Add tools without resorting to git submodule and update variables

omahm:
Dropped adding as submodule as it'll involve more processes (init, update etc) just to install it and the repo isn't updated that often.


johangant (approved):
LGTM!

PR #943, opened 2024-10-09
Dept 929

omahm:
Add d7 database to db service and remove drupal7db service.


johangant (approved):
Haven't tested with the sync-db.sh script but I think it'll be okay as the database key hasn't changed so it'll import to the correct service behind the scenes.

PR #947, opened 2024-10-10
DEPT-925 ~ Remove permissions filter as obsolete in 10.2

ooneill547:


omahm (changes requested):
I wouldn't remove this from composer until after it's been uninstalled.

ooneill547:
I'll create a new branch, it will be quicker than me trying to undo this. Thanks

PR #951, opened 2024-10-14
Dept 946

omahm:
Custom access handler which leverages the account 'Domain Access' field to provide access to Department entities.


ooneill547 (approved):

johangant (approved):
👍

PR #960, opened 2024-10-16
Dept 944

omahm:


johangant (approved):
Worth putting the patch on drupal.org, or referencing it if it's already there in a branch?

omahm:
> Worth putting the patch on drupal.org, or referencing it if it's already there in a branch?

I'm looking to test it on edge first because I had some weird scan differences between edge and local. Once I know it's ok I'll merge it on D.O

PR #969, opened 2024-10-24
Improve psh hostname resolution for depts

johangant:


johangant:
> Won't this assign the psh main url to finance-ni.gov.uk?

This override only happens on branches that aren't `main` or `staging`

omahm:
> > Won't this assign the psh main url to finance-ni.gov.uk?
> 
> This override only happens on branches that aren't `main` or `staging`

Next time I'll maybe look at the entire file 😄

PR #986, opened 2024-11-07
Hostname handling fixes

johangant:


omahm:
Nothing to commit here?

PR #992, opened 2024-11-18
Dept 978

omahm:
activeMigrationDepartments() was using in the body tidy to check against active migrations. I'm leaving it in here just in case we need that type of call anywhere else.


johangant (approved):
:+1:

PR #1003, opened 2024-11-22
Add custom entity access check for media entities

johangant:
Supervisors can now delete entities on domains they are assigned to


omahm (approved):
Yeah I think the hook is the right way to go.

PR #1011, opened 2024-11-29
Preprocess template variables to work around access issues

johangant:
Domain source seems to be a privileged field value (in terms of access) so we can work around some of the access inconsistencies by handling the values in a preprocess hook and setting the values we want consistently that way

### DIFF
--- a/web/themes/custom/nicsdru_dept_theme/templates/content/node--search-metadata.html.twig
+++ b/web/themes/custom/nicsdru_dept_theme/templates/content/node--search-metadata.html.twig
@@ -10,11 +10,11 @@
     {{ snippet }}
   </div>
   <div class="card__meta">
-    {%- if department.id == 'nigov' and domain_source_name %}
-      {%- if 'Agriculture' in domain_source_name %}
+    {%- if department.id == 'nigov' and content_domain_source.name %}
+      {%- if content_domain_source.id == 'daera' %}
         <span class="card__meta-item dept-name">{{ 'DAERA'|t }}</span>
       {% else %}
-        <span class="card__meta-item dept-name">{{ domain_source_name }}</span>
+        <span class="card__meta-item dept-name">{{ content_domain_source.name }}</span>
       {% endif -%}
     {% endif -%}
     {%- if content.field_published_date %}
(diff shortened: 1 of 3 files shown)


web/themes/custom/nicsdru_dept_theme/templates/content/node--search-metadata.html.twig
```diff
@@ -33,3 +28,4 @@
     {% endif -%}
   </div>
 </a>
+{{ content | without('field_summary', 'field_teaser', 'field_domain_source', 'field_published_date') }}
```
johangant, line 31: Looks weird, but you need to do this with more custom templates as it ensures the correct cache tags and set when concluding the rendering pipeline otherwise the render process isn't always sure whether everything's been rendered and may not always correctly finalise the render pipeline

web/themes/custom/nicsdru_dept_theme/templates/content/node--search-metadata.html.twig
```diff
@@ -3,23 +3,18 @@
...
-    {% else %}
-      {{ content.field_summary | field_value }}
-    {% endif -%}
+    {{ snippet }}
   </div>
   <div class="card__meta">
-    {%- if department.id == 'nigov' and department_name %}
-      {%- if 'Agriculture' in department_name %}
+    {%- if department.id == 'nigov' and content_domain_source.name %}
+      {%- if content_domain_source.id == 'daera' %}
```
omahm, line 14: I'm wondering if we should have a utility function to turn dept name into friendly name?

web/themes/custom/nicsdru_dept_theme/templates/content/node--search-metadata.html.twig
johangant: Basically, if you custom render values but your display mode is configured to render these values then you have to either render the via {{ content.thingy }} or finalise with a final {{ content | without('field_names') }} to get the desired result

web/themes/custom/nicsdru_dept_theme/templates/content/node--search-metadata.html.twig
johangant: I'd leave it for now as it's one of the few places the content's underlying source domain is printed but if we have two or more then it would make sense to provide a service to extract and append this to the variables collection so it's there for use in any later templating, a bit like what was done for the current department object

ooneill547 (approved):
It looks complicated, but makes sense

PR #1031, opened 2024-12-12
Dept 1041

omahm:
Fix caching issue when a new topic/subtopic is created by a non-admin users and they use the 'quick add' option on the new node.


ooneill547 (approved):

johangant (approved):
Good spot. I think, in hindsight, cache bins are delimited by `:` like node:123 or media:123 so would make sense to avoid a colon separator in the naming of the bins

PR #1047, opened 2024-12-19
Dept 1079

omahm:


web/themes/custom/nicsdru_dept_theme/images/logo-daera.svg
neilblair, line 1: viewBox attribute is missing and logo will not display correctly.

`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1173 96" width="1173" height="96" preserveAspectRatio="xMinYMin meet">`

instead of

`<svg xmlns="http://www.w3.org/2000/svg" width="1173" height="96" preserveAspectRatio="xMinYMin meet">`

web/themes/custom/nicsdru_dept_theme/images/logo-economy.svg
neilblair, line 1: viewBox attribute is missing and logo does not resize properly on smaller screens.

Should be:
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 468 86" width="468" height="86">`

neilblair (changes requested):

neilblair (commented):
Not sure all the cron timing changes for DEPT-1052 are here. E.g. not seeing config for `*/5 * * * *` for scheduled transitions cron.  When comparing dev to main, can see correct config - https://github.com/dof-dss/nicsdru_dept/compare/main...development

PR #1054, opened 2024-12-23
DEPT-1034 ~ Add image and caption block to layout builder

ooneill547:


omahm:
Are we using Geolocation? I think the module was disabled and marked for removal from composer.

ooneill547:
> Are we using Geolocation? I think the module was disabled and marked for removal from composer.

I wasn't sure and didn't want to remove it if it was needed, I've removed the references again

PR #1061, opened 2025-01-07
Dept 1084

omahm:
Clears the homepage featured cache if a node in that list has it's state changed.


johangant (approved):
Looks good, nice one

PR #1070, opened 2025-01-14
Refine press release saving to match domain access on media entities

johangant:
Ensures that the media entities referenced by the node match the domain access values, but only if it is a press release news type. This ensures that the content and images/media show on the nigov domain as well as the source domain.


web/modules/custom/dept_node/dept_node.module
```diff
@@ -267,9 +272,63 @@ function dept_node_assign_domain_defaults(array $form, FormStateInterface $form_
...
+      if (empty($field_value['selection'])) {
+        continue;
+      }
+
+      // Load the media entity value and set the domain access values for it.
+      foreach ($field_value['selection'] as $key => $value) {
+        $media = Media::load($value['target_id']);
+
+        if ($media instanceof MediaInterface) {
+          $media->set('field_domain_access', $domain_access);
```
omahm, line 326: If it's an existing media entity and has access set on another dept to the current this will overwrite that. Do we need to append this dept to any existing values?

omahm (approved):
LGTM

PR #1081, opened 2025-01-22
DEPT-1072 ~ Reduce font size of captions

ooneill547:


web/themes/custom/nicsdru_dept_theme/src/scss/init/_typography.scss
```diff
@@ -106,6 +106,12 @@ $normal-text-no-margins: (
 );
 
 // Small, medium, large, extra large text (no margins).
+$xs-text: (
+  null: (14px, 20px),
+  xl: (15px, 22px),
+  xxxl: (16px, 24px),
```
neilblair, line 112: Hi Orla, 

Could you change the xl line height to 21px?  It will better maintain vertical rhythm at xl screen sizes.  This sounds mad, but the main font size/line height at xl screen sizes is 18px/27px.  For smaller fonts shown at this screen size, the font size can be anything you want -- but the line height needs to be based on a factor of 18 or 27.   So 21px will work since it is multiple of 3 which is a factor of both 18px and 27px.  Hope that makes sense.

Thanks,
Neil

neilblair (changes requested):

web/themes/custom/nicsdru_dept_theme/src/scss/init/_typography.scss
ooneill547: That makes sense, sorry, that was an oversight

neilblair (approved):

PR #1091, opened 2025-01-29
Hotfix: Update theme version number

neilblair:
Need to update theme number to fix some reported caching problems


omahm (changes requested):
This should have been branched off main.

neilblair:
Yes, seems to be branched of development.  Pretty sure I branched off main.  Will delete and re-do

PR #1097, opened 2025-01-29
DEPT-1127 Update theme version using npm

neilblair:
Every time `npm run build` is run, generate a hash of css/js assets and use as a version number in the theme libraries.yml. This ensures Drupal regenerates hashes for css/js aggregated files and ensures correct cache invalidation behaviour for browsers and edge caches.

I decided go with using a hash instead of incrementing the version number.  If the version number is to follow a major.minor.patch convention, could be problematic incrementing that in a sensible way.  Not sure we really care about the version number that much since we never update it anyway!


ooneill547 (approved):

omahm (approved):
Nice one

PR #1105, opened 2025-02-05
DEPT-1062

johangant:
image

Proposed two col, stacking solution. Only targets the display of the items in the field widget, not in the views based picker form or the media library


omahm:
Looks great

omahm (approved):

PR #1123, opened 2025-02-21
Use entity presave hook to trigger toc removal on books

johangant:
Not enough to use entity update when transitioning between moderation states


omahm (approved):
LGTM

PR #1128, opened 2025-02-25
Dept 1139

omahm:


johangant (approved):
LGTM. Assuming the migrate modules will either remain or be removed at a later point, along with the migration db service

omahm:
> LGTM. Assuming the migrate modules will either remain or be removed at a later point, along with the migration db service

Yeah, can't remove them from composer until prod has had a chance to uninstall.

PR #1133, opened 2025-02-27
DEPT-1154 ~ Patch to fix missing titles on publications, consultation…

omahm:
…s and news


omahm:
fixed in contrib module branch

PR #1138, opened 2025-03-03
Dept 1139

omahm:
remove remaining migration config


composer.json
```diff
@@ -164,8 +164,6 @@
         "drupal/metatag": "^2.1",
         "drupal/metatag_async_widget": "^1.2",
         "drupal/migrate_conditions": "^2.0@beta",
```
johangant, line 166: Remove migrate_conditions module as well?

johangant (approved):
Feels a bit sad seeing all that hard custom migrate work finally being deleted, but that's the nature of the game :)

One query about removing the migrate_conditions module from composer (possibly might need core.extensions.yml checking too) but otherwise GTG

PR #1145, opened 2025-03-13
Dept 1173

omahm:
- Adds revision message when a topic is updated via a non-topic action.
- Added helper method to return the maximum permitted topics per bundle type.


johangant (approved):
LGTM

PR #1156, opened 2025-04-09
Patch antibot and add honeypot module as well

johangant:


johangant:
Closing off as issue queue describes this combination as non-effective in many cases

PR #1204, opened 2025-08-21
Dept 1221

johangant:


johangant:
False positive from git-guardian. It's default http creds for Solr service for local development, not a security risk.

PR #1217, opened 2025-09-26
Fast 404

neilblair:
Return a "friendly" fast 404 for all paths.

Also convert anonymous 403s (which currently get the same fully themed "Page not found" page as 404s) into fast 404s to get:
- crawlers (Google, Bing, bad bots) to eventually drop indexing the URLs
- potential huge reduction in repeated crawl load


johangant (approved):
Great stuff 👍

PR #1222, opened 2025-10-02
Add topics field to events and update hook to set initial value

johangant:
Finding employment topic is to be set. Business area taxonomy term and vocab will be retired at a later point.


omahm (approved):

web/modules/custom/dept_node/dept_node.install
```diff
@@ -1,8 +1,73 @@
...
+    $nodes = $storage->loadMultiple($chunk);
+
+    foreach ($nodes as $node) {
+      // Set field value to single reference for node 409896.
+      $node->set('field_site_topics', [['target_id' => 409896]]);
+      // Clear all values from field_business_area.
+      $node->set('field_business_area', []);
+
+      // Not interested in creating new revisions for this change.
+      $node->setNewRevision(FALSE);
```
omahm, line 59: I pretty sure this will create a revision regardless of this setting due to the workflow module.

PR #1254, opened 2025-11-18
DEPT-1286 ~ Update contrib and patches

omahm:


config/sync/core.extension.yml
```diff
@@ -104,6 +102,7 @@ module:
   jquery_ui: 0
   jquery_ui_autocomplete: 0
   jquery_ui_menu: 0
+  kint: 0
```
johangant: Do we want this in prod?

johangant (commented):
Broadly fine just wondering if we'd do better to put kint config into the local config split profile instead of retaining in general config?

config/sync/core.extension.yml
omahm: yeah that's true although it's a a dev composer install and should ignore it.

johangant (approved):

PR #1309, opened 2026-04-16
DEPT-1304 ~ Revisions config

omahm:


johangant (approved):
Looks good, just that tiny phpstan issue the CI check flagged up. Might need a minor origins release to patch that up, or perhaps ask if @neilblair can override as a one-off thing here if you need the release out for this today

omahm:
> Looks good, just that tiny phpstan issue the CI check flagged up. Might need a minor origins release to patch that up, or perhaps ask if @neilblair can override as a one-off thing here if you need the release out for this today

I've updated Origins and will do another release

PR #1317, opened 2026-04-27
DEPT-1327 Fix revisions overview pagination / missing revisions issue

neilblair:
Pagination issue on revisions overview on some nodes is caused by revisions being filtered from the list because the flag revision_translation_affected is NULL.  Fix is to call $topic_node->setRevisionTranslationAffected(TRUE) before $topic_node->save().  Also setting the revision created time and user id is needed.

See https://share.google/aimode/EV9iO2pdYqMdHzCU1 for full explanation


omahm (approved):
Good catch!

PR #1347, opened 2026-06-10
Fix gallery image captions

johangant:
Render gallery figcaptions from the media caption field, falling back to the image title when no caption is provided. Configure PhotoSwipe to use the media caption field and add the same title fallback for overlay captions.


omahm (approved):
lgtm

PR #1354, opened 2026-07-01
block_related_content: replace node_list with related_content tags

neilblair:
Expands on Johan's existing code for replacing node_list tag on node views with dept_node_list.  For views of related content shown on article pages, the view depends on the topics a node is assigned with. So we can replace node_list with related_content:{topic_nid} and invalidate those tags when a node is changed.  This hopefully delivers highly targeted article cache purges.


johangant (approved):
Looks great 👍

PR #1362, opened 2026-07-06
Dept 806

johangant:
- Use a central class to determine book protection status and control whether node forms permit archiving if there are child pages in the book.
- Introduce a unit test class with data provider and heavily documented readme doc for more clarity on how + why this exists.


omahm (approved):
lgtm

PR #1373, opened 2026-08-04
Dept 1449

omahm:
- New block that fetches news from a rest endpoint.
- Rest response is cached per domain.
- Block has options to change title, more link, number of items and cache duration.


web/modules/custom/dept_ajax_content/dept_ajax_content.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'Department Ajax Content'
+type: module
+description: 'Provides a AJAX displays for frequently changing content to improve caching.'
+core_version_requirement: ^10
```
johangant, line 4: The module references the department.manager service, worth adding a dependency to dept_core module from here?

web/modules/custom/dept_ajax_content/src/Plugin/Block/LatestNewsBlock.php
```diff
@@ -0,0 +1,297 @@
...
+      '#title' => $this->t('"View all news" link URL'),
+      '#description' => $this->t('URL for the "View all news" link shown below the list. Leave empty to hide the link.'),
+      '#default_value' => $this->configuration['view_all_url'],
+      '#maxlength' => 512,
+    ];
+
+    $form['cache_lifetime'] = [
+      '#type' => 'number',
+      '#title' => $this->t('Cache lifetime (seconds)'),
+      '#description' => $this->t('How long to cache API responses. Set to 0 to disable caching.'),
```
johangant: 0 won't disable caching here, it becomes CacheBackendInterface::CACHE_PERMANENT later down the class. So either the form description needs to change to take a value to apply permanent cache (most likely) or the backend implementation for cache values needs to be skipped on a 0 value. But doesn't make sense to remove caching altogether on a feature where we want to improve cache hit ratios :)

web/modules/custom/dept_ajax_content/src/Plugin/Block/LatestNewsBlock.php
```diff
@@ -0,0 +1,297 @@
...
+      $this->logger->error(
+        'Unable to build a valid API URL for the Latest News block. Configure the API URL in the block settings.'
+      );
+      return [];
+    }
+
+    $domain_id = $this->departmentManager->getCurrentDepartment()->id();
+    $cache_id = self::CACHE_ID_PREFIX . md5($url);
+    // Tag matches dept_node_invalidate_latest_news_cache() in dept_node.
+    $cache_tag = 'dept_latest_news:' . $domain_id;
```
johangant: The response gets the `dept_latest_news:{domain}` cache tag and dept_node correctly invalidates that tag. However, the block carries only the unrelated `dept_ajax_content_latest_news` tag at line 203, which is never invalidated. The block will then continue displaying its previously rendered data after the other cache tags are expired.

So I guess some kind of dependency needs to be established between the rendered block and the AJAX response so if one is expired so is the other.

config/sync/block.block.nicsdru_dept_theme_ajaxcontent.yml
```diff
@@ -0,0 +1,29 @@
...
+region: sidebar_second
+weight: -32
+provider: null
+plugin: dept_ajax_content_latest_news
+settings:
+  id: dept_ajax_content_latest_news
+  label: 'Latest News'
+  label_display: visible
+  provider: dept_ajax_content
+  api_url: /api/news/latest
```
johangant, line 22: Can't see what this path is resolving to in the PR. Is it a views path or a missing dept_ajax_content.routing.yml file?

johangant (changes requested):
Is this still a browser/client based thing? There's a few bits to review below but the implementation looks to be a server side rendered block with high cache lifetime and the ticket was looking for a client side request + render in browser to take the pressure off the backend as much as possible.

web/modules/custom/dept_ajax_content/dept_ajax_content.info.yml
omahm: I don't think it's needed as core is always gonna be installed and you wouldn't need to worry about it working as the entire site would fail.

config/sync/block.block.nicsdru_dept_theme_ajaxcontent.yml
omahm: It's the Views Rest endpoint

web/modules/custom/dept_ajax_content/js/ajax-content.js
```diff
@@ -0,0 +1,59 @@
...
+      once('dept-ajax-content', '.ajax-content-block', context).forEach((block) => {
+        const url = block.dataset.ajaxUrl;
+        const count = parseInt(block.dataset.count, 10) || 3;
+        const viewAllUrl = block.dataset.viewAllUrl || null;
+
+        if (!url) {
+          return;
+        }
+
+        fetch(url, { headers: { Accept: 'application/json', 'Cache-Control': 'no-cache', Pragma: 'no-cache' }, cache: 'no-store' })
```
johangant: Cache control headers here need a look:

- cache: 'no-store' tells the browser not to reuse or store the response.
- Cache-Control: no-cache and Pragma: no-cache ask any other caches to revalidate rather than serving their cached copy normally.

With those two in place, we're telling the browser to make a request for the news data on each page even if the backend response is cached by fastly. Ideally, we want the browser to hold onto a version of the response for a sensible period of time to further reduce request rates.

`fetch()` with `headers: { Accept: 'application/json' }` should carry the cache default headers from the backend response. At the moment, it tells the browser it should cache the response locally for 60 seconds which I think is fine for riding out bursts of traffic.

web/modules/custom/dept_ajax_content/js/ajax-content.js
```diff
@@ -0,0 +1,59 @@
...
+
+            const list = document.createElement('ul');
+            list.className = 'ajax-content-list';
+
+            items.slice(0, count).forEach((item) => {
+              const li = document.createElement('li');
+              li.className = 'ajax-content-list__item';
+
+              Object.values(item).forEach((value) => {
+                li.innerHTML += value;
```
johangant: Sent a comment over this bit out of band in Slack

web/modules/custom/dept_ajax_content/js/ajax-content.js
omahm: Good catch.

web/modules/custom/dept_ajax_content/js/ajax-content.js
```diff
@@ -0,0 +1,104 @@
+(function (Drupal, once) {
+  'use strict';
+
+  const BLOCKED_TAGS = ['script', 'style', 'iframe', 'object', 'embed', 'base', 'form'];
+
+  /**
+   * Strips executable content from an HTML string.
```
johangant, line 7: Yep. Let's make it very clear that this HTML must come from a Drupal rendered + filtered response and nothing else. That way we don't need to put in some much safeguarding around untrusted HTML on this part of the process as we always know where it's coming from + offset the risk to Drupal core's XSS protections

johangant (approved):
Looking good now, nice one 👍

PR #1375, opened 2026-08-10
Dept 1472

johangant:


web/modules/custom/dept_topics/dept_topics.module
```diff
@@ -516,6 +529,38 @@ function dept_topics_form_alter(&$form, FormStateInterface $form_state, $form_id
...
+ * Updates topic displays when a node joins or leaves a book hierarchy.
+ */
+function dept_topics_process_book_membership(array $form, FormStateInterface $form_state): void {
+  $values = $form_state->getValues();
+
+  // A node removed from a book becomes eligible for topic display again.
+  if (($form['#form_id'] ?? NULL) === 'book_remove_form') {
+    $entity = \Drupal::routeMatch()->getParameter('node');
+    if ($entity instanceof NodeInterface && $entity->hasField('field_site_topics')) {
+      \Drupal::service('topic.manager')->updateChildDisplayOnTopics($entity);
```
omahm, line 542: One thing to note, this method does not exist in the refactor.

omahm (approved):

web/modules/custom/dept_topics/dept_topics.module
johangant: Better this lower priority work sits a while and waits for the topics refactor to roll out and establish, then I can rework this after. Trying to adapt one or the other in-situ is likely going to end up being messy in the long run.

